PR #24, opened 2020-09-27
Remove unused import from Getting started in README.md

eskombro:
The imported class is not used in the Getting started


djKooks (approved):
Thank you~

PR #30, opened 2020-10-01
basic changes to the project setup

niemannd:
This PR changes three things:

**1) add .editorconfig**

this allows for customization of the editorconfigurations in a unified manner. 
I also ran a reformat with that config in place. Thats the reason for most of the whitespace changes

**2) move code from "meilisearch" package to "com.meilisearch.sdk"**

this change reflects java [package naming conventions](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html).
This will probably make it easier to get the corresponding namespace on maven central and for users to find the library.

**3) move to junit5**

I see no point in starting a project with an old version of junit. This will save us the hassle to migrate to junit5 later on.


.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = tab
+insert_final_newline = false
```
eskombro, line 8:
```suggestion
insert_final_newline = true
```

niemannd:
> I have a question, is it a convention to use `com.meilisearch.sdk` or maybe we should use just `com.meilisearch`?

Its not a convention, but its common. This is useful in case we end up with multiple artifacts (e.g. core, spring integration, etc), so every artifact of the sdk will be located under the "com.meilisearch.sdk" umbrella. 

It also leaves some room in case you want to release some other java libs in the future. Or you decide to rewrite meili in java ;)

eskombro:
I added the `hacktoberfest` label :)

PR #39, opened 2020-10-15
CreateIndex returns Index instance, tested

### LINKED ISSUE #38: CreateIndex should return an Index instance
In the current implementation, `getIndex()` returns an `Index` instance, but `createIndex()` returns the response of the POST request for creating an index. This forces a usage that looks like this:

```java
...
String i = ms.createIndex(indexUid);
Index index = ms.getIndex(indexUid);
index.addDocuments(...)
...
```

In case `createIndex` succeeds, it should return an Index instance which simplifies the usage to 

```java
...
Index index = ms.createIndex(indexUid);
index.addDocuments(...)
...
```

In case of error, this will be handled by https://github.com/meilisearch/meilisearch-java/issues/7

eskombro:
Closes #38

### DIFF
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -32,8 +32,10 @@ public class Client {
 	 * @return Meilisearch API response
 	 * @throws Exception If an error occurs
 	 */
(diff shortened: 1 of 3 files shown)


curquiza (approved):
Easier for the user! Great change! 😁

Issue #22, opened 2020-09-22
Add GetOrCreateIndex method


eskombro:
Gets the index instance or creates a new one if it does not exist

- [ ] Implement `getOrCreate` method
- [ ] Add tests

Follows [this discussion](https://github.com/meilisearch/integration-guides/issues/2)

arjunrc143:
@eskombro Can I give a try on this?

eskombro:
Hello @arjunrc143 

I am very sorry for my late answer!
Indeed, it would be great!

arjunrc143:
> Hello @arjunrc143
> 
> I am very sorry for my late answer!
> Indeed, it would be great!
> 
> Do not hesitate if you need help with anything

Does the master contains breaking changes?

arjunrc143:
https://github.com/meilisearch/meilisearch-java/pull/45

I have added a PR. Please check. I havent updated the integration tests, but I would like to. Can you please help me in setting up environment for integration tests. @eskombro

eskombro:
> Does the master contains breaking changes?

This SDK is still in WIP, so breaking changes are more than common. When we do the release, it will follow the semver system to announce braking changes! But for now, there is no point in trying to be stable :)

eskombro:
> #45
> 
> I have added a PR. Please check. I havent updated the integration tests, but I would like to. Can you please help me in setting up environment for integration tests. @eskombro

Are you in [our Slack](https://slack.meilisearch.com/)? Do not hesitate to contact me there (I'm Samuel) and I will be happy to help you!

We will add some guides soon to this repo, but it is still a bit early to add guides, as everything changes all the time

PR #45, opened 2020-10-27
#22 | Add getOrCreateIndex functionality

arjunrc143:


eskombro:
Hi @arjunrc143 

Thanks for your PR. It is important that in the `getoOrCreateIndex` method we catch the Exception only if the error corresponds to [an index that exists already](https://docs.meilisearch.com/errors/#index_already_exists). Otherwise we shouldn't proceed as we do not necessarily know where the problem is coming from. You can check in the MeiliSearch response if the errorCode corresponds to `index_already_exists`.

This will be updated later when we implement a Custom error handler in #7 

Can you please add this check?

eskombro:
Also, please link the issue. You can add a comment in your pr like this

```
Closes #22 
```

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -107,4 +107,20 @@ public String updateIndex(String uid, String primaryKey) throws Exception {
 	public String deleteIndex(String uid) throws Exception {
 		return this.indexesHandler.delete(uid);
 	}
+
+	/**
+	 * Get single index by uid or if it does not exists, Create index
+	 *
+	 * @param uid        Unique identifier for the index to create
+	 * @param primaryKey The primary key of the documents in that index
+	 * @return Meilisearch API response
```
eskombro, line 116: This function returns an index Instance, not the http response :)

eskombro (changes requested):

src/main/java/com/meilisearch/sdk/Client.java
eskombro, line 116:
```suggestion
	 * @return Index instance
```

eskombro:
```suggestion
	 * @return Index instance
```

arjunrc143:
> Hi @arjunrc143
> 
> Thanks for your PR. It is important that in the `getoOrCreateIndex` method we catch the Exception only if the error corresponds to [an index that exists already](https://docs.meilisearch.com/errors/#index_already_exists). Otherwise we shouldn't proceed as we do not necessarily know where the problem is coming from. You can check in the MeiliSearch response if the errorCode corresponds to `index_already_exists`.
> 
> This will be updated later when we implement a Custom error handler in #7
> 
> Can you please add this check?

Done

arjunrc143:
closes #22

eskombro:
> closes #22

I think this must be done in the description :)

I linked it anyways !

eskombro:
This PR seems good! Just need to add a few integration tests ! For example:

- Use getOrCreateIndex() and check that the index is created
- Use getOrCreateIndex() with an existing index and check that no error is thrown and the index is returned
- Use getOrCreateIndex() 2 or three times in a row...

arjunrc143:
@eskombro I have added the integration tests. Please check.

arjunrc143:
2 integration tests are failing. Will check that.

eskombro:
I checked what's going on, and the problem is that when you get an `index_already_exists` error, the SDK is throwing an IOException with the message 

```
Server returned HTTP response code: 400 for URL: http://localhost:7700/indexes
```

So the check you implemented for the error catching doesn't work. This problem is coming from what `createIndex()` returns in case of failure, and so, not your bad or any propblem in your PR at all actually.

I'll fix this since I'm implementing a basic error handler, and let you know so you can rebase your branch and it should work fine!

Glad we did those tests 👍

arjunrc143:
@eskombro Cool. Let me know once this gets fixed. Thanks!

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -107,4 +107,23 @@ public String updateIndex(String uid, String primaryKey) throws Exception {
...
+	 */
+	public Index getOrCreateIndex(String uid, String primaryKey) throws Exception {
+		try {
+			return this.createIndex(uid, primaryKey);
+		} catch (Exception e) {
+			if(e.getMessage().contains("index_already_exists")) {
+				return this.getIndex(uid);
+			}
+			throw e;
+		}
```
eskombro, line 127:
```suggestion
		try {
			return this.createIndex(uid, primaryKey);
		} catch (MeiliSearchApiException e) {
			if(e.getErrorCode().equals("index_already_exists")) {
				return this.getIndex(uid);
			}
			throw e;
		}
```

eskombro (changes requested):
Hey!

We can already use custom Exceptions, Can you please rebase your branch on master?

I did a suggestion on how to check for the MeiliSearchApiException in this method which is already tested! 

ALSO:

Can you please create a `getOrCreateIndex(String uid)` method without a primaryKey, which calls `getOrCreateIndex(String uid, null)` ?

```java
public Index getOrCreateIndex(String uid) throws Exception {
	return getOrCreateIndex(uid, null);
}
```

`primaryKey` should be optional!

After this PR will be ready to merge! 🎉

arjunrc143:
> Hey @arjunrc143!
> 
> We can already use custom Exceptions, Can you please rebase your branch on master?
> 
> I did a suggestion on your code, on how to check the `MeiliSearchApiException` errorCode in this method which is already tested!
> 
> ALSO:
> 
> Can you please create a `getOrCreateIndex(String uid)` method without a primaryKey, which calls `getOrCreateIndex(String uid, null)` ?
> 
> ```java
> public Index getOrCreateIndex(String uid) throws Exception {
> 	return getOrCreateIndex(uid, null);
> }
> ```
> 
> `primaryKey` should be optional!
> 
> After this PR will be ready to merge! 🎉

@eskombro This is done. Can you please review it?

eskombro (approved):
Thanks! 🎉

PR #74, opened 2020-12-02
Changes due to the implicit index creation

### LINKED ISSUE #76: Implement changes due to implicit index creation
Related to meilisearch/integration-guides#48

- Change the Getting Started as described in the main issue.
- `get_index()` does an HTTP call and is not the main method to use anymore.
- Add the `index()` method
- Add tests for `index()` method
- Update get_or_create_index with the new way of using index()
- Add a `fetchPrimaryKey()` method to `Index` class. The attribute `primaryKey` is not updated when using the `index()` method: the method does not do any HTTP call. Refer to the limitation section in the main issue.

eskombro:
Closes #76

### DIFF
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -68,32 +68,17 @@ public class Client {
 		return meiliSearchIndexList;
 	}
 
-
 	/**
 	 * Create a local reference to an index identified by `uid`, without doing an HTTP call.
-	 * Calling this method doesn't create an index in the MeiliSearch instance, but grants access to all the other methods in the Index class.
+	 * Calling this method doesn't create an index by itself, but grants access to all the other methods in the Index class.
 	 *
 	 * @param uid Unique identifier of the index
 	 * @return Index instance
 	 * @throws Exception If an error occurs
 	 */
 	public Index index(String uid) throws Exception {
-		return this.index(uid, null);
-	}
-
-	/**
-	 * Create a local reference to an index identified by `uid`, without doing an HTTP call.
-	 * Calling this method doesn't create an index by itself, but grants access to all the other methods in the Index class.
-	 *
-	 * @param uid Unique identifier of the index
-	 * @param primaryKey The primary key of the documents in that index
-	 * @return Index instance
-	 * @throws Exception If an error occurs
-	 */
-	public Index index(String uid, String primaryKey) throws Exception {
 		Index index = new Index();
 		index.uid = uid;
-		index.primaryKey = primaryKey;
 		index.setConfig(this.config);
 		return index;
 	}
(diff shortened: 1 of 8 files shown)


src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -68,6 +68,37 @@ public Index createIndex(String uid, String primaryKey) throws Exception {
 		return meiliSearchIndexList;
 	}
 
+
```
curquiza, line 71:
```suggestion
```

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -68,6 +68,37 @@ public Index createIndex(String uid, String primaryKey) throws Exception {
 		return meiliSearchIndexList;
 	}
 
+	
+
+	/**
+	 * Create a local reference to an index identified by `uid`, without doing an HTTP call.
+	 * Calling this method doesn't create an index by itself, but grants access to all the other methods in the Index class.
```
curquiza, line 75:
```suggestion
	 * Calling this method doesn't create an index in the MeiliSearch instance, but grants access to all the other methods in the Index class.
```

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -68,6 +68,37 @@ public Index createIndex(String uid, String primaryKey) throws Exception {
...
+	 *
+	 * @param uid Unique identifier of the index
+	 * @param primaryKey The primary key of the documents in that index
+	 * @return Index instance
+	 * @throws Exception If an error occurs
+	 */
+	public Index index(String uid, String primaryKey) throws Exception {
```
curquiza, line 94: We decide the `index` method should not take a primary key as parameter since it's only a method to define a scope.
We don't use it to do `client.index(uid, primary_key).create()` or `client.index(uid, primary_key).add_documents(...)`, for the moment -> it would mean take the primary key (in RAM) into account and this primary could lead to error/confusion since it could be different from the remote one. That's the decision we took (after a long debate): stay simple for the moment by not including the primary key into `index` method.

src/main/java/com/meilisearch/sdk/Client.java
curquiza: The new line is not necessary, but most of all, don't forget to use the editorconfig plugin to avoid this kind of useless whitespaces 😉

curquiza (changes requested):
Awesome!

- the method `getIndex` should be updated -> this method, that returns an `Index` object if I'm not wrong, should update the `primaryKey` attribute of the `Index` instance with the information returned by MeiliSearch 🙂
- same for `updateIndex` but I opened an issue for this since it also changes the return. So it's not mandatory to change it in this PR 👍  If you do, you can link the issue of course!

curquiza (approved):
awesome! Merge when you want!!! 🚀

eskombro:
Bors merge

PR #83, opened 2020-12-14
Refactor update index method

luis-valdez:
Closes #83 .
Now `updateIndex` returns an instance of Index, and `testUpdateIndexPrimaryKey()` makes sure that this happens.

I was also thinking in including `testUpdateIndexPrimaryKeyIfIndexDoesNotExists()` this test. But I think it's worth having a discussion about it first.
@eskombro and I agree that it should throw a `MeiliSearchApiException`, but what should `updatePrimaryKey` should return when the exception is thrown? Maybe It's worth returning an Optional<Index> for this method?

### DIFF
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -106,8 +106,10 @@ public class Client {
 	 * @return Meilisearch API response
 	 * @throws Exception If an error occurs
 	 */
-	public String updateIndex(String uid, String primaryKey) throws Exception {
-		return this.indexesHandler.updatePrimaryKey(uid, primaryKey);
+	public Index updateIndex(String uid, String primaryKey) throws Exception {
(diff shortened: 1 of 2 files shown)


curquiza (approved):
Looks good to me!
Thanks for your PR @luis-valdez! 😄 
Waiting for the @eskombro before merging

eskombro (approved):
LGTM 👍🏽 

Thanks @luis-valdez for your contribution! 🎉 

And thanks @curquiza for your review/check! 🚀  

About the additional test, maybe this can be discussed in a different issue/PR? As this seems to fulfill the main topic of this issue, and it's better to have atomic commits than mixing up things :)

curquiza:
> About the additional test, maybe this can be discussed in a different issue/PR? As this seems to fulfill the main topic of this issue, and it's better to have atomic commits than mixing up things :)

Agreed, but seems like a good idea! A new issue/PR is appropriate!
Thanks again @luis-valdez 

bors merge

PR #84, opened 2020-12-15
Remove warnings from Json Handling classes

### LINKED ISSUE #72: Remove warnings from Json Handling classes
`JacksonJsonHandlerTest.java` and `GsonJsonHandlerTest.java` generate varargs warnings 

Screenshot 2020-12-02 at 15 34 56

luis-valdez:
Closes #72.

In order to remove the warnings from the classes `GsonJsonHandlerTest` and `JacksonJsonHandlerTest`. I implemented a cast in the `.decode()`  methods last parameter (null) to Class<?>[], this is because the method `.decode()` can return a generic, and in that case, the last argument is expecting a list of types to use with a generic.

TBH I don't know if this is the best approach, so in case it's not please let me know.

### DIFF
--- a/src/test/java/com/meilisearch/sdk/json/GsonJsonHandlerTest.java
+++ b/src/test/java/com/meilisearch/sdk/json/GsonJsonHandlerTest.java
@@ -30,7 +30,7 @@ class GsonJsonHandlerTest {
 	@Test
 	void deserialize() {
 		assertDoesNotThrow(() -> classToTest.decode("{}", Movie.class));
-		assertDoesNotThrow(() -> classToTest.decode("{}", Movie.class, null));
+		assertDoesNotThrow(() -> classToTest.decode("{}", Movie.class, (Class<?>[]) null));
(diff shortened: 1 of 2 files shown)


eskombro (commented):
This seems to fix the problem of the warning quite well. But I am not totally sure I understand the consequences of this, so I want to make sure: 

is this an issue only concerning these tests? I don't think I understand why would we send null `parameters` (3rd argument) to `.decode()` in the first place, and I would be comfortable enough with considering this as undefined behavior, but yet I want to be sure that we are not hiding any potential problem for users later :) But if this approach solves the problem, this is great!

eskombro:
bors try

eskombro:
bors merge

luis-valdez:
I agree with you @eskombro, I'm not quite sure why we were testing with a null parameter in the first place.
As an aditional note, I was using Java 15 and I was receiving no warnings, only when I switched to Java 8 I was able to reproduce the warnings.

eskombro (approved):
Thanks @luis-valdez! 🎉 

Interesting that you didn't get any warning there.

eskombro:
bors merge

PR #88, opened 2020-12-17
Add tests for place holder search

### LINKED ISSUE #15: Test placeholder search
Placeholder search lets the user send a null `query` parameter in order to apply all the ranking rules to the documents in the index without matching them with a specific query, and return the results to the user. It is different to sending an empty string as a query, in which case you won't receive any results. This behavior should be explicitly tested in the SDK.

More info about Placeholder search [here](https://docs.meilisearch.com/guides/main_concepts/search.html#placeholder-search) 

________________________________
UPDATE: 

Placeholder works 👍

Placeholder Search can be used by:
- Using a null query: `q=null`
- Using an empty String: `q=""`

Tests are still missing (both options, null and empty String, should be tested).

luis-valdez:
Closes #15.
In this PR I include to different tests that validate the functionality of place holder search, which consists of making a search a query with an empty string or a null parameter, i.e:
```java
// Both of the following statements return all the documents in the index:
index.search(new SearchRequest(null)); // option 1
index.search(""); // option 2
```

The tests implemented are `testPlaceHolder` and `testPlaceHolderWithLimit`.

I'd appreciate some feedback on the tests because I wasn't sure how atomic the tests should be. If they can be improved in any way, please let me know 😄 !

### DIFF
--- a/src/test/java/com/meilisearch/integration/SearchTest.java
+++ b/src/test/java/com/meilisearch/integration/SearchTest.java
@@ -332,11 +332,12 @@ public class SearchTest extends AbstractIT {
 
 		index.waitForPendingUpdate(updateInfo.getUpdateId());
 
+		String result = index.search("");
 		Results res_gson = jsonGson.decode(
-			index.search(""),
+			result,
 			Results.class
 		);
-		assertEquals(index.search(""), index.search(new SearchRequest(null)));
+		assertEquals(result, index.search(new SearchRequest(null)));
 		assertEquals(20, res_gson.hits.length);
 	}
 
--- a/src/test/java/com/meilisearch/integration/SearchTest.java
+++ b/src/test/java/com/meilisearch/integration/SearchTest.java
@@ -338,12 +338,6 @@ public class SearchTest extends AbstractIT {
 		);
 		assertEquals(index.search(""), index.search(new SearchRequest(null)));
 		assertEquals(20, res_gson.hits.length);
-		assertEquals("419704", res_gson.hits[0].getId());
-		assertEquals("Ad Astra", res_gson.hits[0].getTitle());
-		assertEquals("920", res_gson.hits[10].getId());
-		assertEquals("Cars", res_gson.hits[10].getTitle());
-		assertEquals("420817", res_gson.hits[19].getId());
-		assertEquals("Aladdin", res_gson.hits[19].getTitle());
 	}
 
 	/**
--- a/src/test/java/com/meilisearch/integration/SearchTest.java
+++ b/src/test/java/com/meilisearch/integration/SearchTest.java
@@ -315,5 +315,58 @@ public class SearchTest extends AbstractIT {
 		assertEquals(375, res_gson.hits[0].getMatchesInfo().get("overview").get(2).start);
 		assertEquals(3, res_gson.hits[0].getMatchesInfo().get("overview").get(2).length);
 	}
-	
+	/**
+	 * Test place holder search
+	 */
+	@Test
+	public void testPlaceHolder() throws Exception {
+		String indexUid = "placeHolder";
+		Index index = client.index(indexUid);
+		GsonJsonHandler jsonGson = new GsonJsonHandler();
+
+		TestData<Movie> testData = this.getTestData(MOVIES_INDEX, Movie.class);
+		UpdateStatus updateInfo = jsonGson.decode(
+			index.addDocuments(testData.getRaw()),
+			UpdateStatus.class
+		);
+
+		index.waitForPendingUpdate(updateInfo.getUpdateId());
+
+		Results res_gson = jsonGson.decode(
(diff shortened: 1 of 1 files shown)


src/test/java/com/meilisearch/integration/SearchTest.java
```diff
@@ -315,5 +315,58 @@ public void testSearchMatches() throws Exception {
...
+			UpdateStatus.class
+		);
+
+		index.waitForPendingUpdate(updateInfo.getUpdateId());
+
+		Results res_gson = jsonGson.decode(
+			index.search(""),
+			Results.class
+		);
+		assertEquals(index.search(""), index.search(new SearchRequest(null)));
```
eskombro, line 339: I would suggest storing the results of `index.search("")` in a Sting variable and then use it in both places (L336 and L339). This won't probably add a noticeable difference now, but seems to me to be a more consistent way of knowing the values are the same always, and also avoid 1 call to the server :)

src/test/java/com/meilisearch/integration/SearchTest.java
```diff
@@ -315,5 +315,58 @@ public void testSearchMatches() throws Exception {
...
+		assertEquals(20, res_gson.hits.length);
+		assertEquals("419704", res_gson.hits[0].getId());
+		assertEquals("Ad Astra", res_gson.hits[0].getTitle());
+		assertEquals("920", res_gson.hits[10].getId());
+		assertEquals("Cars", res_gson.hits[10].getTitle());
+		assertEquals("420817", res_gson.hits[19].getId());
+		assertEquals("Aladdin", res_gson.hits[19].getTitle());
```
eskombro, line 346: Checking these exact hardcoded values may seem to make sense at first sight, but it can introduce a few problems that may pop up in the future. As MeiliSearch Core continues to evolve, changes in the way the MeiliSearch algorithm works may end up breaking these tests, even if the SDK functionality is still OK (imagine for example that de default order of the ranking rules changes, and so, the order of the results is not the same. But the Placeholder search is still working, right? well according to this test, no. This is the complexity that we are introducing here). So from my POV, we could make this more "generic". We need to know that call works, that it returns results, that we have the default amount of them (20)... but checking the specific ID is always risky. I know we do it sometimes in some other tests, but in this case doesn't seem to be really necessary, WDYT?

eskombro (commented):
Good work, thanks! I added two comments, let me know what you think about this :)

eskombro:
Also, I was thinking that it would be interesting to add a test checking that both ways of using placeholder search produce the same result (null and empty string). Am I going too far and this is more of a core test than a SDK test @curquiza ?

curquiza:
> Also, I was thinking that it would be interesting to add a test checking that both ways of using placeholder search produce the same result (null and empty string). Am I going too far and this is more of a core test than a SDK test @curquiza ?

The first test already does it if I'm not wrong:

```java
assertEquals(index.search(""), index.search(new SearchRequest(null)));
```

For me this is enough 🙂

src/test/java/com/meilisearch/integration/SearchTest.java
luis-valdez: Good catch @eskombro! I didn't notice I was actually making one unnecessary call to the server. I'll implement the changes.

src/test/java/com/meilisearch/integration/SearchTest.java
luis-valdez: Tbh I was a little confused on how I had to prove that placeholder was working, but now that you mention it, it makes a lot of sense testing it without the hardcoded values.
Checking only the amount of results seems like a good approach. Thank you for the feedback!

luis-valdez:
> > Also, I was thinking that it would be interesting to add a test checking that both ways of using placeholder search produce the same result (null and empty string). Am I going too far and this is more of a core test than a SDK test @curquiza ?
> 
> The first test already does it if I'm not wrong:
> 
> ```java
> assertEquals(index.search(""), index.search(new SearchRequest(null)));
> ```
> 
> For me this is enough 🙂

That's correct. I thought about doing two separate tests, one for the empty string and another one for the null value.
In one hand having two separate tests would make them more atomic, but in the other hand a lot of code would be duplicated.
What do you guys think?

curquiza:
> That's correct. I thought about doing two separate tests, one for the empty string and another one for the null value.
In one hand having two separate tests would make them more atomic, but in the other hand a lot of code would be duplicated.
What do you guys think?

I would rather one test, so let it as it is if @eskombro is ok. If we find out any trouble with this choice in the future, we will change! 🙂

eskombro (approved):
Good, I think this is enough for testing Placeholder functionality and raise an alert if something is broken. Thanks @luis-valdez !

eskombro:
bors try

eskombro:
bors merge

PR #91, opened 2020-12-20
fix lombok version in build.gradle

niemannd:
had some problems to get lombok running with intellij 2020.3 due to the version mismatch.

PR #92, opened 2020-12-20
implement ServiceTemplate 

niemannd:
⚠️ contains the changes of #91 

This PR introduces the "ServiceTemplate". While the HttpClient wraps how http calls are done and the JsonHandler wraps the (de-)serialization logic, the ServiceTemplate defines how these two abstractions work together. 

The GenericServiceTemplate defines the default behaviour for our Implementations. 
Make the Request -> deserialize the response -> return Response

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,9 +59,9 @@ dependencies {
 
 	// Lombok
 	compileOnly 'org.projectlombok:lombok:1.18.16'
-	annotationProcessor 'org.projectlombok:lombok:1.18.12'
-	testCompileOnly 'org.projectlombok:lombok:1.18.12'
-	testAnnotationProcessor 'org.projectlombok:lombok:1.18.12'
+	annotationProcessor 'org.projectlombok:lombok:1.18.16'
+	testCompileOnly 'org.projectlombok:lombok:1.18.16'
+	testAnnotationProcessor 'org.projectlombok:lombok:1.18.16'
 }
 
 task buildJar(type: Jar) {
--- a/src/main/java/com/meilisearch/sdk/GenericServiceTemplate.java
+++ b/src/main/java/com/meilisearch/sdk/GenericServiceTemplate.java
@@ -1,0 +1,107 @@
+package com.meilisearch.sdk;
+
+import com.meilisearch.sdk.exceptions.MeiliSearchRuntimeException;
+import com.meilisearch.sdk.http.AbstractHttpClient;
+import com.meilisearch.sdk.http.HttpClient;
+import com.meilisearch.sdk.http.factory.RequestFactory;
+import com.meilisearch.sdk.http.request.HttpRequest;
+import com.meilisearch.sdk.http.response.HttpResponse;
+import com.meilisearch.sdk.json.JsonHandler;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+public class GenericServiceTemplate implements ServiceTemplate {
+	private final AbstractHttpClient client;
+	private final JsonHandler processor;
+	private final RequestFactory requestFactory;
+
+	/**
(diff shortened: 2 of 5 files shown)


curquiza:
Hello @niemannd, thanks for both of your PRs. @eskombro is on holidays for the moment but will come back soon and will review all the PRs 🙂
We will release the package right after all the merges.
Thanks again for your involvement!

eskombro:
HI @niemannd !

I am currently reviewing this PR! I am struggling a bit in implementing it with this level of abstraction. For example, I am using this code to create an index:

```java
public static void main(String[] args) throws Exception {

        GenericServiceTemplate ss = new GenericServiceTemplate(
            new DefaultHttpClient(new Config("http://localhost:7700", "masterKey")),
            new GsonJsonHandler(),
            new BasicRequestFactory()
        );

        HashMap<String, String> headerMap = new HashMap();
        headerMap.put("X-Meili-API-Key", "masterKey");

        Index index = ss.execute(
            ss.getRequestFactory().create(
                HttpMethod.POST,
                "/indexes",
                headerMap,
                "{\"uid\":\"books\"}"
            ),
            Index.class
        );

        System.out.println(index.getUid());
    }
```

The index is created, but `index.getUid()` returns `null`. I assumed that the `execute()` method was deserializing the server response into the index instance. Is there something I'm missing? Shouldn't this return "books"?

eskombro:
I would also like to ask what would be the expected implementation of this ServiceTemplate in the SDK?

niemannd:
> The index is created, but index.getUid() returns null. I assumed that the execute() method was deserializing the server response into the Index instance, AKA, index. Is there something I'm missing? Shouldn't this return "books"?

I can only reproduce this behaviour if the index already exists. The error handling is one weakspot of this setup. I have to think about how to do that better. Maybe simply throw an exception. 

> I would also like to ask what would be the expected implementation of this ServiceTemplate in the SDK?
See https://github.com/niemannd/meilisearch-java-forked/blob/da78870c0a6440c5d24023bf876479afd307f48c/src/main/java/com/meilisearch/sdk/api/index/IndexesHandler.java#L75
This way the API itself does not rely on any other specific component.

eskombro:
Thanks @niemannd after checking indeed this problem only comes when index is already created, and indeed we need to improve the error handling in his implementation. I have been checking your implementation of the service templates in your repo, and I think I have more visibility about the big picture now! I will have a discussion with the integration team and will come back to you very soon, but thanks a lot for this work! I may come with a question or two :)

eskombro (commented):
Thanks @niemannd after checking indeed this problem only comes when index is already created, and indeed we need to improve the error handling in his implementation. I have been checking your implementation of the service templates in your repo, and I think I have more visibility about the big picture now! I will have a discussion with the integration team and will come back to you very soon, but thanks a lot for this work! I may come with a question or two :)

eskombro:
bors try

eskombro:
Hi @niemannd !

Thanks for this contribution. After having reviewed and tested this approach of Service Template, I think that it is actually improving the quality of the SDK. From my point of view, this simplifies a lot the implementation of the API and takes out the complexity of it, and this is great because this is the part of the SDK that needs more maintenance and where contributors come first. This level of abstraction solves several problems and gives the user a high level of freedom in choosing its own dependencies and environment. Nonetheless, I think we will be trying to document everything to make this accessible to someone who doesn't know the project because the complexity is not disappearing, but being abstracted and moved into the implementation of the templates themselves, HTTP Client, and JSON handlers.

I created #105 for the implementation (description can be upgraded, for this was just a 'placeholder' issue), do you want to be assigned to this?

Thanks a lot for this PR and for making this SDK evolve! 

And sorry for the late answer :)

eskombro:
bors merge

eskombro (approved):

PR #95, opened 2020-12-24
Implement setting methods

### LINKED ISSUE #93: Implement settings methods
There are 3 main methods that are missing in this package. If I'm wrong, feel free to close this issue, but I don't find anything when I search for `settings` in this repository.

- `getSettings` that corresponds to this route: https://docs.meilisearch.com/references/settings.html#get-settings
- `updateSettings` that corresponds to this route: https://docs.meilisearch.com/references/settings.html#update-settings
- `resetSettings` that corresponds to this route: https://docs.meilisearch.com/references/settings.html#reset-settings

The `updateSettings` route should be able to work even if all the settings (`rankingRules`, `distinctAttribute`, `synonyms`...) are not all passed. A similar implementation to the `SearchRequest` class with all the `setXXX` methods for the `Settings` class could be useful: https://github.com/meilisearch/meilisearch-java/blob/2b042ada803cd207730b85db4e10157a9ed3d923/src/main/java/com/meilisearch/sdk/SearchRequest.java.

Steps:
- [ ] Add the `Settings` class with all the needed methods
- [ ] Add the 3 settings methods (get, update, reset)
- [ ] Add tests

Feel free to ask any question if something is not clear to you 🙂

luis-valdez:
#  1. General Info
Closes #93.
This PR implements 3 methods:
- `getSettings`
- `updateSettings`
- `resetSettings`
Refers to https://docs.meilisearch.com/references/settings.html

# 2. Add Settings class
This class has the methods that make the http calls ( `getSettings`, `updateSettings`, `resetSettings`). Uses `meiliSearchHttpRequest` to do so.

# 3. Add SettingsRequest class
I added this class to follow the example of the [SearchRequest class]( https://github.com/meilisearch/meilisearch-java/blob/2b042ada803cd207730b85db4e10157a9ed3d923/src/main/java/com/meilisearch/sdk/SearchRequest.java). It is the main data structure involved in settings.
It is helpful for two reasons:
1. It helps us deserialize JSON objects incoming form the server
2. It allows us to easily modify with `setters` our settings to use `updateSettings`

# 4. Modify Index class
I implemented the following changes:
1. Add settings as a field
2. Include settings in the constructor
3. Implemented the 3 main methods, that subsequently call the methods from the `Settings` class

# 5. Add SettingsTest class
I added 4 tests to validate that everything was working fine.
- `testGetSettings`
-  `testUpdateSettingsRankingRules`
- `testUpdateSettingsSynonyms`
- `testResetSettings`

There are 2 main things I'd like to remark on this part.
1. I used a `Thread.sleep(1000)` to ensure that the `updateSettings` was completed, the 1000 was completely arbitrary and I'd like to know if it's okay.
2. Maybe we don't need `testUpdateSettingsSynonyms` if we already have `testUpdateSettingsRankingRules`.

***
As always I'd really appreciate your feedback. And if you think this can be improved in any way please let me know.

Happy Holidays 😄 🎉 !


src/main/java/com/meilisearch/sdk/Index.java
```diff
@@ -160,6 +164,40 @@ public String search(SearchRequest sr) throws Exception {
...
+	 *
+	 * @return settings of a given uid as String
+	 * @throws Exception if something goes wrong
+	 */
+	public String getSettings() throws Exception {
+		return this.settings.getSettings(this.uid);
+	}
+
+	/**
+	 * Updates the settings in index
```
eskombro, line 179:
```suggestion
	 * Updates the settings of the index
```

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -0,0 +1,183 @@
...
+	static void cleanMeiliSearch() {
+		cleanup();
+	}
+
+	/**
+	 * Test get settings from an index by uid
+	 */
+	@Test
+	public void testGetSettings() throws Exception {
+		String indexUid = "updateSettings";
```
eskombro, line 41:
```suggestion
		String indexUid = "getSettings";
```

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+		assertEquals(2, newSettings.getSynonyms().size());
+
+		index.resetSettings();
+		Thread.sleep(1000);
+		SettingsRequest settingsAfterReset = jsonGson.decode(
+			index.getSettings(),
+			SettingsRequest.class
+		);
+		assertEquals(0, settingsAfterReset.getSynonyms().size());
```
eskombro, line 180: For now, the default is 0, but I don't know if this could ever change. What if, instead of comparing it to 0, you compare that the size is the same it returned before modifying the settings? (Meaning, comparing it to the length you got on L155)

eskombro (commented):
Thanks for this amazing work @luis-valdez ! Once again :)

I added some little suggestions that from my POV can improve some things a little, those are just that, suggestions.

By reviewing this PR, which seems to works very well as expected, I was thinking about 3 things/questions that I would like to share:

1. The `sleep(1000)` can be replaced by using the `waitForPendingUpdate(updateId)` method. This method actually checks when the update has been processed (or failed) instead of waiting for an arbitrary amount of time, which risks:
- being too long and making tests last longer than necessary, or ...
- not being long enough and make them fail :)
So using this method should make the tests more consistent, I think.

2. As for now, Settings are returned as a String, but wouldn't it be better to use the same class that you called `SettingsRequest` to deserialize them?
As in the last PR we merged (with Dumps), I think that the currently named `Settings` class can be considered as a "settings handler" and the currently named 'SettingsRequest` could be just a `Settings` class. The `getSettings()` method could return an already unserialized object, an instance of this class. I think this would improve the user experience. What do you think?

Let me know if my explanation is not clear enough, it happens from time to time :)

3. I was wondering, is there any reason to use Settings outside of an Index Handler? Or maybe it can fit inside there? I have no strong opinion on this, just thought it was worth considering

Anyways, great job, thanks again :)

luis-valdez:
Hello @eskombro, great feedback as always!

First of all, I believe I covered all the minor changes you suggested, which are:
- Comments
- String uid values in tests

I really agree with what you said about
>As for now, Settings are returned as a String, but wouldn't it be better to use the same class that you called SettingsRequest to deserialize them?
As in the last PR we merged (with Dumps), I think that the currently named Settings class can be considered as a "settings handler" and the currently named SettingsRequest could be just a Settings class. The getSettings() method could return an already unserialized object, an instance of this class. I think this would improve the user experience. What do you think?
>
I think the first architecture I implemented was a little bit confusing and a little uncomfortable for the user.

I implemented the following changes:
- `getSettings` is returning a Settings object
-  `updateSettings` and `resetSettings` are now returning an UpdateStatus Object
-  `sleep(1000` is changed for `waitForPendingUpdate`
-  What was `Settings` is now `SettingsHandler`
-  What was `SettingsRequest` is now `Settings`

And for now I think that's it.

I'm really scratching my head about this one:
>I was wondering, is there any reason to use Settings outside of an Index Handler? Or maybe it can fit inside there? I have no strong opinion on this, just thought it was worth considering
>

As I see it right now we have two alternatives:
**Option A)**
Right now `Index` has a `SettingsHandler` to help that particular index manage its own Settings, which makes sense to me because when you want to use the Settings methods, you ask to that particular index to update/get/reset its own settings, like this:
```java
Index index = client.index(indexUid);
Settings settings = index.getSettings();
```
I think this option has better readability.

**Option B**
The other alternative, if I understood correctly, would be to get rid of `SettingsHandler` and include its methods in `IndexesHandler` and subsequently use the methods in the `Client`. It would look something like this:
```java
Index index = client.index(indexUid);
Settings settings = client.getIndexSettings(index.getUid);
```
Which I think has worse readability.
But as you said, there is no reason to use `SettingsHandler` outside of `IndexesHandler`. 

I'm really struggling to make a decision here, so I'd really appreciate more feedback on this 😄

eskombro (approved):
Hi @luis-valdez ! Thanks a lot for all of that work!!!

After reading your feedback and thinking about it, it seems that the way you have implemented it is way better than the second option. Looking at the example makes it more than clear. It seems way more logical, usable, and readable. But thanks for taking it into consideration!

This PR looks very good to me, amazing contribution 🎉

eskombro:
bors try

eskombro:
Everything seems good. Let me know if from your Point of view this is ready to merge! 🎉

curquiza (approved):
LGTM 👌 Thank you so much @luis-valdez for your involvement!
I let you merge with bors @eskombro when you think this is ok

eskombro:
Thanks, waiting for @luis-valdez opinion.

And happy new year (you too Bors, you too) :)

luis-valdez:
If there are no conflicts, I'd say go ahead and merge 🚀

I wish you too guys a happy new year! @eskombro @curquiza (and bors too...) 🎉

eskombro:
Gracias 🎉 
Bors merge

PR #89, opened 2020-12-18
Implement dumps

luis-valdez:
## General info
Closes #35.
This PR implements the dumps route of the MeiliSearch API. 
For more details about dumps: https://docs.meilisearch.com/references/dump.html

### Implement dump creation route

 The new class `Dump` works as a wrapper for `MeiliSearchRequestHttp` and handles the creation of the dump with the method `createDump`. This method returns a Dump object which contains:
- `uid` that is the unique identifier of the dump
- `status` which indicates the status of the dump at the moment of the call

The method `createDump` is finally called by the `Client` class in a method of its own, that also returns an instance of Dump.

###  Implement get dump status route
Very similar to the previous point, it uses the new `Dump` class in coordination with the `Client` class to make an HTTP GET request and retrieve the current information of the dump. It uses the method `getStatusByUid(String uid)` in the Dump class and the `getDumpStatus(String uid)` in the Client class.
An example return value of `getDumpStatus`
```
{"uid":"20201218-023314939","status":"done"}
```
 
 ### Add tests
To validate the previous 2 points, two new tests were implemented in  the `ClientTest` class. This tests are `testGetDumpStatus` and `testCreateDump`.
1. `testCreateDump` creates a dump and validates if it has a status of `in_progress`
2. `testGetDumpStatus` creates a dump and gets its uid. Then it validates if it can retrieve the status using `getStatusByUid`

 ### Add code samples
In this part I wasn't sure where should I include the code samples. But here they are:
```
import com.meilisearch.sdk.Client;
import com.meilisearch.sdk.Config;
import com.meilisearch.sdk.Index;
import com.meilisearch.sdk.Dump;

class TestMeiliSearch {
  public static void main(String[] args) throws Exception {

    final String documents = "["
      + "{\"book_id\": 123, \"title\": \"Pride and Prejudice\"},"
      + "{\"book_id\": 456, \"title\": \"Le Petit Prince\"},"
      + "{\"book_id\": 1, \"title\": \"Alice In Wonderland\"},"
      + "{\"book_id\": 1344, \"title\": \"The Hobbit\"},"
      + "{\"book_id\": 4, \"title\": \"Harry Potter and the Half-Blood Prince\"},"
      + "{\"book_id\": 2, \"title\": \"The Hitchhiker\'s Guide to the Galaxy\"}"
      + "]";

    Client client = new Client(new Config("http://localhost:7700", "masterKey"));
    
    // An index is where the documents are stored.
    Index index = client.index("books");

[...]

Issue #19, opened 2020-09-21
Choose a linter and integrate it to CI and docs


eskombro:
- [ ] Choose a well maintained and known Linter for Java
- [ ] Run the linter for the whole SDK code and make a Pull Request
- [ ] Add a task in the CI [GitHub Actions] that runs automatically the linter (Lint will be required for any merge/contribution afterwards)

Do not hesitate to contact us if you have any doubts or questions!

AmiralBl3ndic:
I started looking for a linter and [`error-prone`](https://errorprone.info/index) seems to be a good solution.
There is no official gradle plugin for it though, we will have to rely upon [the community-developed plugin](https://github.com/tbroyer/gradle-errorprone-plugin)

eskombro:
It seems interesting. A pity that it doesn't have a code formatting tool. The plugin seems good.

What do you think if we use https://github.com/google/google-java-format?

AmiralBl3ndic:
@eskombro I heard some friends talking about using [`prettier-java`](https://github.com/jhipster/prettier-java) for code formatting. I haven't personally tried it, but it seems like a good option to achieve code formatting in Java. Just be advised they also told me that it requires a bit configuration and this relies on each user having a node environment setup to use with `npm` or `yarn`.

eskombro:
Sorry for the late answer @AmiralBl3ndic, and thanks for the suggestion!

I changed team at MeiliSearch, and probably @alallema will be addressing this topic and issue when she gets some time :)

Thanks for all your help, again 🎉

ppshobi:
We could close this issue as #187 is merged, and we are using spotless, which is a no dependency linter. Also a CI step has been added

PR #96, opened 2020-12-30
Fix bad styling

curquiza:
Minor fixes about styling introduced by https://github.com/meilisearch/meilisearch-java/pull/89.


eskombro (approved):
Thanks!

We really should find a solution to #19 

:)

curquiza:
bors merge

Issue #68, opened 2020-12-01
Generic class for search results


eskombro:
In the current state of the Java SDK, the return value of the search method is a String, which needs to be deserialized by the user after performing a search.

This SDK should provide a generic class capable of deserializing MeiliSearch results. Search method should return an instance of this object, with MeiliSearch's results already deserialized.

djKooks:
@eskombro I think I could work on it. 

For case which returning value does not have fixed form, how do you think of returning value as `Map` object?

eskombro:
Amazing @djKooks !

Are you talking about the "hits" field? if that's the case, I think a Map or a HashMap Are indeed the best options :)

djKooks:
@eskombro I've create PR for this. It only includes very raw commits, just want to share whether it is the one you're intending. Please check 🙏

PR #108, opened 2021-01-16
Response with generic class for `Search`

djKooks:
Hello
This is PR for #68 . Please look on 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ Output:
 
 ## 🤖 Compatibility with MeiliSearch
 
-This package only guarantees the compatibility with the [version v0.18.0 of MeiliSearch](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.18.0).
+This package only guarantees the compatibility with the [version v0.18.1 of MeiliSearch](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.18.1).
 
 ## 💡 Learn More
 
--- a/README.md
+++ b/README.md
@@ -108,16 +108,13 @@ import com.meilisearch.sdk.model.SearchResult;
 
 // MeiliSearch is typo-tolerant:
 SearchResult results = index.search("harry pottre");
-System.out.println(results.getOffset());
-System.out.println(results.getLimit());
-System.out.println(results.getQuery());
+System.out.println(results);
 ```
 
-the result will be:
+- Output:
+
 ```
-0
-20
-harry pottre
+SearchResult(hits=[{book_id=4.0, title=Harry Potter and the Half-Blood Prince}], offset=0, limit=20, nbHits=1, exhaustiveNbHits=false, facetsDistribution=null, exhaustiveFacetsCount=false, processingTimeMs=3, query=harry pottre)
 ```
 
 #### Custom Search <!-- omit in toc -->
@@ -138,7 +135,8 @@ String results = index.search(
 System.out.println(results.getHits());
 ```
 
-Output:
+- Output:
+
 ```json
 [{
   "book_id":1,
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-  <img src="https://raw.githubusercontent.com/meilisearch/integration-guides/master/assets/logos/logo.svg" alt="MeiliSearch Java" width="200" height="200" />
+  <img src="https://raw.githubusercontent.com/meilisearch/integration-guides/main/assets/logos/logo.svg" alt="MeiliSearch Java" width="200" height="200" />
 </p>
 
 <h1 align="center">MeiliSearch Java</h1>
@@ -16,7 +16,7 @@
 <p align="center">
   <a href="https://maven-badges.herokuapp.com/maven-central/com.meilisearch.sdk/meilisearch-java"><img src="https://maven-badges.herokuapp.com/maven-central/com.meilisearch.sdk/meilisearch-java/badge.svg" alt="Version"></a>
   <a href="https://github.com/meilisearch/meilisearch-java/actions"><img src="https://github.com/meilisearch/meilisearch-java/workflows/Tests/badge.svg" alt="Tests"></a>
-  <a href="https://github.com/meilisearch/meilisearch-java/blob/master/LICENSE"><img src="https://img.shields.io/badge/license-MIT-informational" alt="License"></a>
+  <a href="https://github.com/meilisearch/meilisearch-java/blob/main/LICENSE"><img src="https://img.shields.io/badge/license-MIT-informational" alt="License"></a>
   <a href="https://app.bors.tech/repositories/29365"><img src="https://bors.tech/images/badge_small.svg" alt="Bors enabled"></a>
 </p>
 
@@ -37,7 +37,7 @@
 
 ## 📖 Documentation
 
-See our [Documentation](https://docs.meilisearch.com/guides/introduction/quick_start_guide.html) or our [API References](https://docs.meilisearch.com/references/).
+See our [Documentation](https://docs.meilisearch.com/learn/tutorials/getting_started.html) or our [API References](https://docs.meilisearch.com/reference/api/).
 
 
 ## 🔧 Installation
@@ -97,7 +97,7 @@ class TestMeiliSearch {
 }
 ```
 
-With the `updateId`, you can check the status (`enqueued`, `processed` or `failed`) of your documents addition using the [update endpoint](https://docs.meilisearch.com/references/updates.html#get-an-update-status).
+With the `updateId`, you can check the status (`enqueued`, `processed` or `failed`) of your documents addition using the [update endpoint](https://docs.meilisearch.com/reference/api/updates.html#get-an-update-status).
 
 #### Basic Search <!-- omit in toc -->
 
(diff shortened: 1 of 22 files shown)


eskombro (commented):
Hey @djKooks 

Thanks for starting this PR. It is indeed the idea behind #68 (or at least what I had in mind). 🎉 

The implementation in PHP SDK seem to be a great one if some inspiration is needed at any point:

https://github.com/meilisearch/meilisearch-php/blob/master/src/Search/SearchResult.php

If I understand, your code's goal is to create a class in which the server response can be deserialized when a search request is performed. In this way, the user can access the SearchResult object and its attributes/methods immediately. Looks perfect 👍🏽

djKooks:
@eskombro update the PR. Please give feedback if there's any problem 🙏

src/main/java/com/meilisearch/sdk/model/SearchResult.java
```diff
@@ -0,0 +1,36 @@
...
+
+	@Getter
+	boolean exhaustiveNbHits;
+
+	@Getter
+	int processingTimeMs;
+
+	@Getter
+	String query;
+}
```
curquiza, line 36: Sorry if I miss something, but I think `exhaustiveFacetsCount` and `facetsDistribution` are missing and are really useful 😊

djKooks: @curquiza thanks~
Also, how do you think about name? Is `SearchResult` okay, or how about `SearchResponse`?
Please let me know if you have suggestion...

curquiza: I would rather `SearchResult` 🙂 
"Response" refers more to the raw body returned by MeiliSearch, and not the class, from my POV.

src/test/java/com/meilisearch/integration/SearchTest.java
```diff
@@ -217,13 +183,10 @@ public void testSearchHighlight() throws Exception {
...
-			index.search(searchRequest),
-			Results.class
-		);
-		assertEquals(20, res_gson.hits.length);
-		assertTrue(res_gson.hits[0].getFormatted().getOverview().contains("<em>"));
-		assertTrue(res_gson.hits[0].getFormatted().getOverview().contains("</em>"));
+
+		SearchResult searchResult = index.search(searchRequest);
+
+		assertEquals(20, searchResult.getHits().size());
```
djKooks, line 189: @curquiza @eskombro 

Currently I've implemented `hits` as `Array<HashMap<String, Object>>`, but not sure whether this is the best way.
Maybe it could be better to use `Array<Object>` to allow end user to serialize the format inside array more easily...

How do you think?

eskombro (changes requested):
Hi @djKooks !

Thanks a lot for this PR. The implementation looks good to me. I added a few comments on the tests that were modified, only because I feel that some test that were removed where important to check that we are sending the right request to MeiliSearch. Can you check it?

Other than that, this looks good to me! Thanks!!!! 🎉

src/test/java/com/meilisearch/integration/SearchTest.java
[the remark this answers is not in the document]
djKooks: @eskombro it can be, but seems `Movie` class is only for test, and `_formated` is the parameter only for this use case. 

As I commented above, that is why I'm thinking of using `Array<Object>` instead of `Array<HashMap>` to allow user to setup serialization...but not sure which is the better way.

src/test/java/com/meilisearch/integration/SearchTest.java
```diff
@@ -299,21 +258,10 @@ public void testSearchMatches() throws Exception {
...
-		assertEquals(52, res_gson.hits[0].getMatchesInfo().get("overview").get(0).start);
-		assertEquals(3, res_gson.hits[0].getMatchesInfo().get("overview").get(0).length);
-		assertEquals(214, res_gson.hits[0].getMatchesInfo().get("overview").get(1).start);
-		assertEquals(3, res_gson.hits[0].getMatchesInfo().get("overview").get(1).length);
-		assertEquals(375, res_gson.hits[0].getMatchesInfo().get("overview").get(2).start);
-		assertEquals(3, res_gson.hits[0].getMatchesInfo().get("overview").get(2).length);
+		SearchRequest searchRequest = new SearchRequest("and").setMatches(true);
+		SearchResult searchResult = index.search(searchRequest);
+
+		assertEquals(20, searchResult.getHits().size());
```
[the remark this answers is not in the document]
djKooks: @eskombro same as above comments...

eskombro (commented):
Hi @djKooks !

Concerning the choice of `ArrayList<HashMap<String, Object>>` or `ArrayList<Object>`: From my POV, I the ease of access to any element is a great plus: 

```Java
SearchResult result = index.search("t");
ArrayList<HashMap<String, Object>> hits = result.getHits();
for (int i = 0; i < hits.size(); i++) {
    System.out.println(hits.get(i).get("title"));
}
```

> Maybe it could be better to use Array<Object> to allow end user to serialize the format inside array more easily...


What way of deserialization do you have in mind for this case?

eskombro (commented):
Also @djKooks we would like to expose to the user the `SearchResult` functionality for it's ease of use, but not force it. For this, we can implement the same solution as the PHP SDK which is offering a `rawSearch()` method which returns the results in the same way as they were implemented before, and a `search()` method returning the `SearchResult` class implemented by you in this PR. 

EXAMPLE: https://github.com/meilisearch/meilisearch-php/blob/22333b1e302d96853a45639c9e75474a3e3928e0/src/Endpoints/Indexes.php#L146-L173

The `search()` method calls `rawSearch()` and transforms the response, serializing it into `SearchResult`.

The user can choose the option that fits better his usecase/preference.

What do you think?

djKooks:
@eskombro @curquiza I've added `rawSearch` for returning raw string result, as you mentioned.
Could you check? I'll update document if it seems okay.

eskombro:
Hey @djKooks 

This implementation looks great! Thanks! I am not sure I understood what you meant by "updating document". But other than that this LGTM! 🎉

eskombro:
Bors try

djKooks:
@eskombro `updating document` means writing comment about new functions:
https://github.com/meilisearch/meilisearch-java/pull/108/files#diff-2e36c6e8b505832fb749583632d24f72a107625b3720ee007524ceada0d86fd3R157-R163

thanks for review~

eskombro:
Perfect @djKooks, that would be great!

djKooks:
@eskombro @curquiza document is also updated.
Please confirm and approve if it's okay(or you can merge it directly) 🙏

curquiza:
@eskombro will review this on Monday, thanks @djKooks 🙂

eskombro (changes requested):
Hi @djKooks !

Sorry to bother you again with this! The documentation additions in-code are great! 

Nonetheless, we wouldn't want to introduce this kind of changes on the README directly, because all of the SDK READMEs have a specific structure, and that helps us keep coherence and automate maintenance, etc. To document `rawSearch()` method in [PHP SDK](https://github.com/meilisearch/meilisearch-php) (for example) there is a [Wiki](https://github.com/meilisearch/meilisearch-php/wiki/Search) and this wiki is [linked on the README](https://github.com/meilisearch/meilisearch-php/blob/main/README.md#custom-search-). We could try this approach, but this isn't blocking to merge your PR, so we can do this afterwards.

Is it possible to take out the README changes introduced in this PR for now? I'll merge it as soon as possible !

Thanks again!

eskombro:
I created #128 to tackle this separately and finally merge your contribution  :)

djKooks:
@eskombro thanks~updated it.

README.md
```diff
@@ -101,28 +101,23 @@ With the `updateId`, you can check the status (`enqueued`, `processed` or `faile
...
 ```java
+import com.meilisearch.sdk.model.SearchResult;
+
 // MeiliSearch is typo-tolerant:
-String results = index.search("harry pottre");
-System.out.println(results);
+SearchResult results = index.search("harry pottre");
+System.out.println(results.getOffset());
+System.out.println(results.getLimit());
+System.out.println(results.getQuery());
```
eskombro, line 113:
```suggestion
System.out.println(results);
```

README.md
```diff
@@ -101,28 +101,23 @@ With the `updateId`, you can check the status (`enqueued`, `processed` or `faile
...
-	"nbHits": 1,
-	"exhaustiveNbHits": false,
-	"processingTimeMs": 2,
-	"query": "harry pottre"
-}
+the result will be:
+```
+0
+20
+harry pottre
```
eskombro, line 120:
```suggestion
- Output:
```
SearchResult(hits=[{book_id=4.0, title=Harry Potter and the Half-Blood Prince}], offset=0, limit=20, nbHits=1, exhaustiveNbHits=false, facetsDistribution=null, exhaustiveFacetsCount=false, processingTimeMs=3, query=harry pottre)
```

File: README.md [hunk 6]
```diff
@@ -101,28 +101,23 @@ With the `updateId`, you can check the status (`enqueued`, `processed` or `faile
...
-		"title": "Harry Potter and the Half-Blood Prince"
-	}],
-	"offset": 0,
-	"limit": 20,
-	"nbHits": 1,
-	"exhaustiveNbHits": false,
-	"processingTimeMs": 2,
-	"query": "harry pottre"
-}
+the result will be:
```
eskombro, line 116:
```suggestion
- Output:
```

README.md
eskombro, line 120:
```suggestion
SearchResult(hits=[{book_id=4.0, title=Harry Potter and the Half-Blood Prince}], offset=0, limit=20, nbHits=1, exhaustiveNbHits=false, facetsDistribution=null, exhaustiveFacetsCount=false, processingTimeMs=3, query=harry pottre)
```

djKooks: @eskombro fixed~please check 🙏

eskombro (approved):
🎉

eskombro:
Thanks @djKooks !

Bors try

eskombro:
Bors merge

PR #120, opened 2021-02-07
improve ErrorHandling

niemannd:
This PR improves the ErrorHandling of the GenericServiceTemplate. Instead of ignoring an error, we throw it as an RuntimeException. This way the user can decide if he wants to handle it or just ignore it.

### DIFF
--- a/src/test/java/com/meilisearch/integration/ExceptionsTest.java
+++ b/src/test/java/com/meilisearch/integration/ExceptionsTest.java
@@ -2,12 +2,9 @@ package com.meilisearch.integration;
 
 import com.meilisearch.integration.classes.AbstractIT;
 import com.meilisearch.sdk.Index;
-import com.meilisearch.sdk.exceptions.MeiliSearchException;
+import com.meilisearch.sdk.exceptions.APIError;
 import com.meilisearch.sdk.exceptions.MeiliSearchApiException;
-import org.junit.jupiter.api.AfterAll;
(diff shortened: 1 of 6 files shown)


src/test/java/com/meilisearch/integration/ExceptionsTest.java
```diff
@@ -54,11 +45,12 @@ public void testErrorSerializeAndGetters() throws Exception {
 	 * Test MeiliSearchApiException is thrown on MeiliSearch bad request
 	 */
 	@Test
+	@Disabled
```
eskombro, line 48:
```suggestion
```

eskombro (dismissed):
Hey @niemannd !

Thanks for this Error handler improvement! The only modification I would like to do is removing the `@disabled` for `testMeiliSearchApiExceptionBadRequest` that you probably forgot (but the test is passing btw). I did a suggestion for this in the review, can you please fix it? Otherwise, LVGTM :)

eskombro:
bors try

niemannd:
The problem with cutting big rewrites into smaller PRs is that you sometimes get stuff like that `@Disabled` in there.
I removed it. It would be fixed in the following PRs anyway.

eskombro:
Ok, sorry about that, I thought it would be better to review and merge one by one because it becomes quite big and I don't want to miss the details :)

eskombro (approved):
👌🏽

eskombro:
Bors merge

PR #239, opened 2021-10-11
Added getRawIndex and getRawIndexList methods

### LINKED ISSUE #238: Add method to get the raw index information
⚠️ This issue is generated, it means the examples and the namings do not necessarily correspond to the language of this repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue.

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue in the repo.

Related to https://github.com/meilisearch/integration-guides/issues/120

----

This SDK provides a method to get an index (`getIndex`) and this method returns an instance of `Index`. Some users want to get the raw version of the index, I mean the exact JSON response returned by the MeiliSearch server, not an instance.

Some SDKs already provide a way to get this raw index information but this is not convenient, example in PHP:

```php
$client->index('movies')->fetchRawInfo();
```

We need to add the `getRawIndex` method, accessible from the `Client` instance, to make it more convient.

Prototype:

```
getRawIndex(string indexUID) -> object/struct/dict of the MeiliSearch response
```

Ex in PHP:
```php
$client->getRawIndex('movies');
```

TODO:
- [x] Add `getRawIndex()` method to the `Client` instance without duplicating the code (if `fetchRawInfo()` already exists for example)
- [x] Add tests

vishnugt:
Fixes #238

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -1,0 +1,403 @@
+# This code-samples file is used by the MeiliSearch documentation
+# Every example written here will be automatically fetched by
+# the documentation on build
+# You can read more on https://github.com/meilisearch/documentation/tree/master/.vuepress/code-samples
+---
+get_one_index_1: |-
+  client.getIndex("movies");
+list_all_indexes_1: |-
+  client.getAllIndexes();
+create_an_index_1: |-
+  client.createIndex("movies", "movie_id");
+update_an_index_1: |-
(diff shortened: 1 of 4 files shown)


vishnugt:
I was thinking if I should return `String` or a `JsonObject` but since existing methods like `deleteIndex` returns `String`, I also returned String.

I wasn't sure if the issue also talked about `getRawIndexList` method, I added it anyway. :)
cc: @alallema

alallema:
bors merge

PR #241, opened 2021-10-13
Remove todo comments in code samples

curquiza:
The documentation will not display the `Java` tab if it cannot find the associated text for the code sample. I would rather no tab being displayed rather than a `// todo`
WDYT?


alallema:
I think you are right, better nothing than `//todo` but anyway we can submit it to the doc before they are all complete. Otherwise, it will throw an error on the script in the documentation no?

alallema:
@curquiza,
Also there is some: `//Not yet implemented`. Did we remove then too or maybe exchange for `Not existing` ?

PR #248, opened 2021-10-16
Add delete index if exists method

ezienecker:
ISSUE-231: Add delete index if exists method to handle index not found error


ezienecker:
@alallema thanks for the review. I implement the review remarks

PR #255, opened 2021-10-19
Update README.md by replacing `movie_id` by `id`

curquiza:
According to https://github.com/meilisearch/meilisearch-java/issues/229, we should not use `movie_id` but `id`


alallema:
bors merge

PR #345, opened 2022-03-02
Adding generateTenantToken method to the client

alallema:
## Tenant tokens

Introduction of the new method `generateTenantToken`  in order to facilitate the generation of the tenant token.

Related to:
- this issue: https://github.com/meilisearch/meilisearch/issues/1991
- this spec: https://github.com/meilisearch/specifications/pull/89


src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -242,4 +248,58 @@ public Key createKey(Key options) throws Exception {
...
+            throw new MeiliSearchException(
+                    "An api key is required in the client or should be passed as an argument.");
+        }
+        if (options.getExpiresAt() != null
+                && Date.from(Instant.now()).after((Date) options.getExpiresAt())) {
+            throw new MeiliSearchException("The date expiresAt should be in the future.");
+        }
+        if (searchRules == null) {
+            throw new MeiliSearchException(
+                    "The search_rules field is mandatory and should be defined.");
```
brunoocasali, line 282:
```suggestion
                    "The searchRules field is mandatory and should be defined.");
```

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -242,4 +248,58 @@ public Key createKey(Key options) throws Exception {
...
+        }
+
+        String secret;
+        if (options.getApiKey() == null) {
+            secret = this.config.apiKey;
+        } else {
+            secret = (String) options.getApiKey();
```
brunoocasali, line 289: `getApiKey` could respond with something different than `String`?

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -242,4 +248,58 @@ public Key createKey(Key options) throws Exception {
...
+        }
+
+        // Encrypte the Key
```
brunoocasali, line 292:
```suggestion
        // Encrypt the key
```

src/main/java/com/meilisearch/sdk/Key.java
```diff
@@ -5,8 +5,6 @@
 
 /** MeiliSearch response for a Key */
 public class Key {
-    // This field has been set to "" pending resolution of this issue
-    // https://github.com/meilisearch/meilisearch/issues/2116
```
brunoocasali, line 9: Can you look if there are other occurrences of this?

src/main/java/com/meilisearch/sdk/TenantTokenOptions.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public String getApiKey() {
+        return apiKey;
+    }
+
+    public void setApiKey(String apiKey) {
+        this.apiKey = apiKey;
+    }
+
+    public Date getExpiresAt() {
```
brunoocasali, line 21: This `Date` can handle time as well?

src/test/java/com/meilisearch/integration/TenantTokenTest.java
```diff
@@ -0,0 +1,244 @@
...
+        Client privateClient = new Client(new Config("http://localhost:7700", key.getKey()));
+
+        Map<String, Object> rules = new HashMap<String, Object>();
+        rules.put("*", new HashMap<String, Object>());
+
+        String jwtToken = privateClient.generateTenantToken(rules);
+
+        Client tokenClient = new Client(new Config("http://localhost:7700", jwtToken));
+
+        SearchResult searchResult = tokenClient.index(indexUid).search("");
```
brunoocasali, line 58: Since you threw an `Exception` if something goes wrong in the search request you can just ensure there is no Exception after the request :)

src/test/java/com/meilisearch/integration/TenantTokenTest.java
```diff
@@ -0,0 +1,244 @@
...
+        String indexUid = "GenerateTokenWithExpiresAt";
+        Index index = createEmptyIndex(indexUid);
+        Key key = getPrivateKey();
+
+        Client privateClient = new Client(new Config("http://localhost:7700", key.getKey()));
+
+        Map<String, Object> rules = new HashMap<String, Object>();
+        rules.put("*", new HashMap<String, Object>());
+
+        Instant instant = Instant.now();
```
brunoocasali, line 131: This is only supported by Java 8, right?

src/test/java/com/meilisearch/integration/TenantTokenTest.java
```diff
@@ -0,0 +1,244 @@
...
+        Key key = getPrivateKey();
+
+        Client privateClient = new Client(new Config("http://localhost:7700", key.getKey()));
+
+        assertThrows(MeiliSearchException.class, () -> privateClient.generateTenantToken(null));
+    }
+
+    /** Test Create Tenant Token with bad expireation date */
+    @Test
+    public void testGenerateTenantTokenWithBadExpiresAt() throws Exception {
```
brunoocasali, line 208: Could you add a test about UTC as well?

brunoocasali (commented):
Good Job @alallema ☕

src/test/java/com/meilisearch/integration/TenantTokenTest.java
alallema: Yes, it is the reason I just use it in the tests and not in the code. Do you think it could be a problem?

src/main/java/com/meilisearch/sdk/Client.java
alallema: Yes, I added it because I had problems with the linter but it seems to be fixed.

src/main/java/com/meilisearch/sdk/Key.java
alallema: Your mean about this issue in this SDK or other issues in this SDK or this issue in the other SDK?

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -242,4 +248,58 @@ public Key createKey(Key options) throws Exception {
...
+     *     be used. - expiresAt: A DateTime when the key will expire. Note that if an expiresAt
+     *     value is included it should be in UTC time.
+     * @return String containing the tenant token
+     * @throws MeiliSearchException if an error occurs
+     */
+    public String generateTenantToken(Map<String, Object> searchRules, TenantTokenOptions options)
+            throws MeiliSearchException {
+        // Validate all fields
+        if ((options.getApiKey() == null || options.getApiKey() == "")
+                && this.config.apiKey == "") {
```
brunoocasali, line 272: `this.config.apiKey` can be also `null`?

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -242,4 +248,58 @@ public Key createKey(Key options) throws Exception {
...
+                && Date.from(Instant.now()).after((Date) options.getExpiresAt())) {
+            throw new MeiliSearchException("The date expiresAt should be in the future.");
+        }
+        if (searchRules == null) {
+            throw new MeiliSearchException(
+                    "The searchRules field is mandatory and should be defined.");
+        }
+
+        String secret;
+        if (options.getApiKey() == null) {
```
brunoocasali, line 286: @alallema if the config.apiKey is valid like `"88392che832ebh"` and the result of `options.getApiKey()` is `""` (empty string), your condition will not fall in the `MeiliSearchException` in the line 273. And will not fall in the 287 as well. Your result will be a `secret = "";`

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -242,4 +248,58 @@ public Key createKey(Key options) throws Exception {
...
+        }
+
+        // Encrypt the key
+        Algorithm algorithm = Algorithm.HMAC256(secret);
+
+        // Create JWT
+        String jwtToken =
+                JWT.create()
+                        .withClaim("searchRules", searchRules)
+                        .withClaim("apiKeyPrefix", secret.substring(0, 8))
```
brunoocasali, line 299: It is possible to have a `StringIndexOutOfBoundsException` here if this condition **Either starting or ending index is greater than the total number of characters present in the string.** happens

src/main/java/com/meilisearch/sdk/Key.java
brunoocasali: Nope, I saw that you fixed the issue here https://github.com/meilisearch/meilisearch-java/pull/344 and then removed all the comments like this one, from the files, but you missed this. My question is, there are other comments about this subject lost somewhere?

src/test/java/com/meilisearch/integration/TenantTokenTest.java
brunoocasali: I think the Instant is a great addition, but since we support java 7, we need to be able to run the test suite with java 7. We should work to create a process for this kind of situation. So I think the better is to remove the non-supported features in Java 7.

src/test/java/com/meilisearch/integration/classes/AbstractIT.java
```diff
@@ -69,6 +69,17 @@ public void loadResource(String fileName) throws IOException {
         testData.put(fileName, new TestData<>(rawData, movies));
     }
 
+    public Key getPrivateKey() throws Exception {
```
brunoocasali, line 72: Don't you need this kind of key in other tests? (just making sure that this is not a duplicated code somewhere)

brunoocasali (commented):
I left some points for you @alallema good job!

src/main/java/com/meilisearch/sdk/Key.java
alallema: I hope not this time 😄

src/test/java/com/meilisearch/integration/classes/AbstractIT.java
alallema: No, I just need it in the TokenTokenTest did you think it's better to add it only in `src/test/java/com/meilisearch/integration/TenantTokenTest.java`?

brunoocasali: Not really, could be useful in the future :)

src/main/java/com/meilisearch/sdk/Client.java
alallema: Yes but only if you create your client like:
```
      Client client = new Client(new Config("http://localhost:7700", null));
```
Instead of:
```
      Client client = new Client(new Config("http://localhost:7700"));
```
I will ad the verification anyway

src/main/java/com/meilisearch/sdk/Client.java
alallema: Except if the user put the `masterKey` in it it shouldn't happen all the keys are generated by `Meilisearch`.

src/test/java/com/meilisearch/integration/TenantTokenTest.java
alallema: I'm not sure to know what you think about it?

src/test/java/com/meilisearch/integration/TenantTokenTest.java
alallema: Remove!

src/main/java/com/meilisearch/sdk/TenantTokenOptions.java
alallema: What do you mean?

brunoocasali: If you want to deal with `Date` in ruby you can use [`Time`](https://ruby-doc.org/core-3.1.0/Time.html) or [`Date`](https://ruby-doc.org/stdlib-3.1.1/libdoc/date/rdoc/Date.html), but if you want to deal with both you cannot use `Date`. My question is related to the `time` part of a date and if this Date class behaves like Ruby's or not 😄

src/test/java/com/meilisearch/integration/TenantTokenTest.java
brunoocasali: I've added validations to expiresAt to check if they come as UTC value both in dart and ruby, I don't know if this is overkill (because the user can have problems with that).

src/main/java/com/meilisearch/sdk/Client.java
brunoocasali: I suggest you add that validation since we cannot prevent the user from adding that key to the method invocation. If it happens, the user will face an exception he does not expects, even before getting the token :)

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -242,4 +248,59 @@ public Key createKey(Key options) throws Exception {
...
+            throw new MeiliSearchException(
+                    "An api key is required in the client or should be passed as an argument.");
+        }
+        if (searchRules == null) {
+            throw new MeiliSearchException(
+                    "The searchRules field is mandatory and should be defined.");
+        }
+
+        String secret;
+        if (options.getApiKey() == null || options.getApiKey() == "") {
```
brunoocasali, line 287: This validation repeats 4 times in this code, I'm not sure if it happens outside of this scope as well. But we could create a small function to help clean that out!

```java
private boolean isNullOrEmpty(String str) {
   return str == null || str.trim().length() == 0;
}
```

Or use `com.google.common.base.Strings.isNullOrEmpty(String string)` (but I'm not sure how guava is used in the meilisearch-java, since I just found it in the deps but not in the code).

src/test/java/com/meilisearch/integration/TenantTokenTest.java
alallema: Alright, get it. Yes, I'm not really comfortable adding this verification I'm afraid it could create errors on the user side depending on how he prefers to use this function. I think a note in the definition of the function is enough no?

src/main/java/com/meilisearch/sdk/TenantTokenOptions.java
alallema: It could be quite similar because the Class `Date` encapsulates the `Date` and `Time` but the `Time` is just used for (hour, minutes, second, millisecond ...) I'm not really sure you can use `Time` for a `Date` if you need a `Date`.

src/main/java/com/meilisearch/sdk/Client.java
alallema: I just find 2 of it and I modify the code to check the length

bidoubiwa (approved):
LGTM 🔥🔥🔥

PR #370, opened 2022-04-21
Fix test due to changes v0.27.0

alallema:
Changes in the request result of a search query due to a new implementation of `typo-tolerance`.
Until now `typo` on the first letter is now counting as 2 typos

### DIFF
--- a/src/test/java/com/meilisearch/integration/SearchTest.java
+++ b/src/test/java/com/meilisearch/integration/SearchTest.java
@@ -270,7 +270,7 @@ public class SearchTest extends AbstractIT {
 
         SearchResult searchResult = index.search(searchRequest);
 
-        assertEquals(3, searchResult.getHits().size());
+        assertEquals(1, searchResult.getHits().size());
         assertNotNull(searchResult.getFacetsDistribution());
     }
 


src/test/java/com/meilisearch/integration/SearchTest.java
```diff
@@ -270,7 +270,7 @@ public void testSearchFacetsDistribution() throws Exception {
 
         SearchResult searchResult = index.search(searchRequest);
 
-        assertEquals(3, searchResult.getHits().size());
+        assertEquals(1, searchResult.getHits().size());
```
bidoubiwa, line 273: Why do you receive fewer hits?

alallema: I think it's because of the `typo tolerance` feature.
I was receiving 3 results for the query `knight` including 2 with `night`
Now I receive only 1 result with `knight`

alallema: Do you think it's not an expected behavior?

bidoubiwa: I'm not sure, could you open an issue in `meilisearch/meilisearch` ? Since we are removing acceptable candidates

alallema: It is an expected behavior 😅 https://github.com/meilisearch/meilisearch/issues/2348

brunoocasali (approved):
🥩 LGTM!

PR #371, opened 2022-04-21
Add typo tolerance settings

alallema:
This PR introduces the new setting: [typoTolerance](https://github.com/meilisearch/specifications/pull/117)

new methods: 
`index.getTypoTolerance()`
`index.updateTypoTolerance(params)`
`index.resetTypoTolerance()`

### DIFF
--- a/src/main/java/com/meilisearch/sdk/SettingsHandler.java
+++ b/src/main/java/com/meilisearch/sdk/SettingsHandler.java
@@ -390,4 +390,50 @@ public class SettingsHandler {
                 meilisearchHttpRequest.delete("/indexes/" + uid + "/settings/distinct-attribute"),
                 Task.class);
     }
+
+    /**
+     * Gets the typo tolerance settings of a given index Refer
+     * https://docs.meilisearch.com/reference/api/typo-tolerance.html#get-typo-tolerance
+     *
+     * @param uid Index identifier
+     * @return a Map that contains all typo tolerance and their associated words
+     * @throws Exception if an error occurs
+     */
+    public Map<String, String[]> getTypoToleranceSettings(String uid) throws Exception {
+        return this.gson.<Map<String, String[]>>fromJson(
+                meilisearchHttpRequest.get("/indexes/" + uid + "/settings/typo-tolerance"),
+                Map.class);
+    }
+
+    /**
+     * Updates the typo tolerance settings of a given index Refer
+     * https://docs.meilisearch.com/reference/api/typo-tolerance.html#update-typo-tolerance
+     *
+     * @param uid Index identifier
+     * @param typoTolerance a Map that contains the new typo tolerance settings
+     * @return Task instance
+     * @throws Exception if an error occurs
+     */
+    public Task updateTypoToleranceSettings(String uid, Map<String, String[]> typoTolerance)
+            throws Exception {
+        String typoToleranceAsJson = gson.toJson(typoTolerance);
+        return this.gson.fromJson(
+                meilisearchHttpRequest.post(
+                        "/indexes/" + uid + "/settings/typo-tolerance", typoToleranceAsJson),
+                Task.class);
+    }
+
+    /**
+     * Resets the typo tolerance settings of a given index Refer
+     * https://docs.meilisearch.com/reference/api/typo-tolerance.html#reset-typo-tolerance
+     *
+     * @param uid Index identifier
+     * @return Task instance
+     * @throws Exception if an error occurs
+     */
+    public Task resetTypoToleranceSettings(String uid) throws Exception {
+        return this.gson.fromJson(
+                meilisearchHttpRequest.delete("/indexes/" + uid + "/settings/typo-tolerance"),
+                Task.class);
+    }
 }
--- a/src/main/java/com/meilisearch/sdk/Settings.java
+++ b/src/main/java/com/meilisearch/sdk/Settings.java
@@ -19,6 +19,7 @@ public class Settings {
     @Getter @Setter private String[] searchableAttributes;
     @Getter @Setter private String[] displayedAttributes;
     @Getter @Setter private String[] sortableAttributes;
+    @Getter @Setter private HashMap<String, String[]> typoTolerance;
 
     /** Empty SettingsRequest constructor */
     public Settings() {}
@@ -53,6 +54,9 @@ public class Settings {
         }
         if (this.getSortableAttributes() != null) {
             jsonObject.put("sortableAttributes", this.getSortableAttributes());
+        }
+        if (this.getTypoTolerance() != null) {
+            jsonObject.put("typoTolerance", this.getTypoTolerance());
         }
         return jsonObject.toString();
     }


src/main/java/com/meilisearch/sdk/Index.java
```diff
@@ -567,6 +567,40 @@ public Task resetDistinctAttributeSettings() throws Exception {
         return this.settingsHandler.resetDistinctAttributeSettings(this.uid);
     }
 
+    /**
+     * Get the typo tolerance field of an index.
+     * https://docs.meilisearch.com/reference/api/distinct_attribute.html#get-distinct-attribute
```
brunoocasali, line 572: Hey, this link is not correct :)

src/main/java/com/meilisearch/sdk/TypoTolerance.java
```diff
@@ -0,0 +1,60 @@
...
+    protected String[] disableOnWords;
+    protected String[] disableOnAttributes;
+
+    /**
+     * Method to return the JSONObject of the TypoTolerance Setting
+     *
+     * @return JSONObject of the TypoTolerance Setting object
+     */
+    public JSONObject toJson() {
+        JSONObject jsonObject = new JSONObject();
```
brunoocasali, line 21: @alallema I think you can just use: `gson.toJson(this);` don't you?

This will convert complex java objects to a json string.

src/main/java/com/meilisearch/sdk/TypoTolerance.java
```diff
@@ -0,0 +1,60 @@
...
+import org.json.JSONObject;
+
+public class TypoTolerance {
+
+    public TypoTolerance() {}
+
+    protected boolean enabled = false;
+    protected HashMap<String, Integer> minWordSizeForTypos;
+    protected String[] disableOnWords;
+    protected String[] disableOnAttributes;
```
brunoocasali, line 13: In the class `public class Settings {` you used `@Getter` and `@Setter`, this can be applied here as well, right?

src/main/java/com/meilisearch/sdk/api/index/Settings.java
```diff
@@ -18,6 +19,7 @@ public class Settings {
     private String[] searchableAttributes;
     private String[] displayedAttributes;
     private String[] sortableAttributes;
+    private TypoTolerance typoTolerance;
```
brunoocasali, line 22: `@Getter` and `@Setter` here as well?

brunoocasali (commented):
@alallema I didn't review the tests yet :)

src/main/java/com/meilisearch/sdk/TypoTolerance.java
alallema: Yes, it depends on the class some have `getter` and `setter wrote as a function, and other has `@Getter @Setter`.I'm not sure to understand exactly why but I try to keep the same as the other class file like `Key` or `Details` class. Maybe we should create an issue to harmonize everything?

brunoocasali: I think you should definitely create the issue, but also do the changes (for the files you touched in this branch) now.

I was looking for some references to this particular topic: "wait to do all the refactor at once or do incrementally", and I found this https://meiert.com/en/blog/boyscout-code/, a good reading!

src/main/java/com/meilisearch/sdk/SettingsHandler.java
```diff
@@ -390,4 +390,50 @@ public Task resetDistinctAttributeSettings(String uid) throws Exception {
                 meilisearchHttpRequest.delete("/indexes/" + uid + "/settings/distinct-attribute"),
                 Task.class);
     }
+
+    /**
+     * Gets the typo tolerance settings of a given index Refer
+     * https://docs.meilisearch.com/reference/api/typo-tolerance.html#get-typo-tolerance
```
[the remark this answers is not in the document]
alallema: It will when the doc will be release, normally 😄

src/main/java/com/meilisearch/sdk/TypoTolerance.java
alallema: c:
```bash
"typoTolerance":"{\"enabled\":true,\"disableOnWords\":[\"and\"],\"disableOnAttributes\":[\"title\"]}","displayedAttributes":["*"]}
```
And cause an error:
```bash
 MeiliSearchApiException{error=APIError{message='Json deserialize error: invalid type: string "{\"enabled\":true,\"disableOnWords\":[\"and\"],\"disableOnAttributes\":[\"title\"]}", expected struct TypoSettings at line 1 column 285', code='bad_request', type='invalid_request', link='https://docs.meilisearch.com/errors#bad_request'}}
```

src/main/java/com/meilisearch/sdk/Index.java
```diff
@@ -567,6 +567,40 @@ public Task resetDistinctAttributeSettings() throws Exception {
...
+        return this.settingsHandler.getTypoToleranceSettings(this.uid);
+    }
+
+    /**
+     * Update the typo tolerance field of an index.
+     * https://docs.meilisearch.com/reference/api/distinct_attribute.html#update-distinct-attribute
+     *
+     * @param typoTolerance A TypoTolerance instance
+     * @return Task instance
+     * @throws Exception if an error occurs
```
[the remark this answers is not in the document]
alallema: It throw an Exception like:
```bash
Exception in thread "main" MeiliSearchApiException{error=APIError{message='Index `movies` not found.', code='index_not_found', type='invalid_request', link='https://docs.meilisearch.com/errors#index_not_found'}}
```

src/main/java/com/meilisearch/sdk/SettingsHandler.java
```diff
@@ -390,4 +390,50 @@ public Task resetDistinctAttributeSettings(String uid) throws Exception {
...
+     *
+     * @param uid Index identifier
+     * @param typoTolerance a TypoTolerance instance that contains the new typo tolerance settings
+     * @return Task instance
+     * @throws Exception if an error occurs
+     */
+    public Task updateTypoToleranceSettings(String uid, TypoTolerance typoTolerance)
+            throws Exception {
+        String typoToleranceAsJson = gson.toJson(typoTolerance);
+        return this.gson.fromJson(
```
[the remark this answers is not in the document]
alallema: Did you mean I had to add `<Task>` to this?
```java
    public Task updateSettings(String uid, Settings settings) throws Exception {
        return this.gson.<Task>fromJson(
                meilisearchHttpRequest.post(
                        "/indexes/" + uid + "/settings", settings.getUpdateQuery()),
                Task.class);
    }
```

src/main/java/com/meilisearch/sdk/TypoTolerance.java
alallema: I open an issue about it #379

src/main/java/com/meilisearch/sdk/SettingsHandler.java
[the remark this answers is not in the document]
brunoocasali: Yes, the first option, because I saw it on line 403.

src/main/java/com/meilisearch/sdk/Index.java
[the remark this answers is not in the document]
brunoocasali: Yes, that's I'm talking about, why not defining like this: `throws MeiliSearchApiException` ?

src/main/java/com/meilisearch/sdk/TypoTolerance.java
brunoocasali: But I think it was the correct behavior, now the problem is here:

```java
// src/main/java/com/meilisearch/sdk/Settings.java:58

jsonObject.put("typoTolerance", this.getTypoTolerance().toJson());
```

Since I don't know much about this gson API, we could do this:

```java
    return (JsonObject) gson.toJsonTree(this);
```

brunoocasali: Based on this https://stackoverflow.com/a/47211859/2649707

src/main/java/com/meilisearch/sdk/api/index/Settings.java
```diff
@@ -1,88 +1,27 @@
...
 public class Settings {
-    private HashMap<String, String[]> synonyms;
-    private String[] stopWords;
-    private String[] rankingRules;
-    private String[] filterableAttributes;
-    private String distinctAttribute;
-    private String[] searchableAttributes;
-    private String[] displayedAttributes;
-    private String[] sortableAttributes;
+    @Getter @Setter private HashMap<String, String[]> synonyms;
```
brunoocasali, line 15: Much better to the 👀 ! 😅

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -101,6 +104,27 @@ public void testUpdateSettingsSort() throws Exception {
         assertEquals(2, newSettings.getSortableAttributes().length);
     }
 
+    @Test
+    @DisplayName("Test update settings changing the typo tolerance")
+    public void testUpdateSettingsTypoTolerance() throws Exception {
+        Index index = createIndex("updateSettingsSynonyms");
```
brunoocasali, line 110:
```suggestion
        Index index = createIndex("updateSettingsTypoTolerance");
```

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -101,6 +104,27 @@ public void testUpdateSettingsSort() throws Exception {
...
+        typoTolerance.setEnabled(true);
+        typoTolerance.setDisableOnWords(new String[] {"and"});
+        typoTolerance.setDisableOnAttributes(new String[] {"title"});
+        settings.setTypoTolerance(typoTolerance);
+
+        index.waitForTask(index.updateSettings(settings).getUid());
+
+        Settings newSettings = index.getSettings();
+
+        assertEquals(1, newSettings.getTypoTolerance().getDisableOnWords().length);
```
brunoocasali, line 123: I think you should assert by the value here because if the `getDisableOnWords` responds with a `["other"]` your test will not break.

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -101,6 +104,27 @@ public void testUpdateSettingsSort() throws Exception {
...
+        assertEquals(1, newSettings.getTypoTolerance().getDisableOnAttributes().length);
+        assertEquals(true, newSettings.getTypoTolerance().isEnabled());
```
brunoocasali, line 125: Why not use the [assertTrue](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertTrue-boolean-java.util.function.Supplier-) for this cases?

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -176,18 +201,19 @@ public void testResetRankingRuleSettings() throws Exception {
         index.waitForTask(index.updateRankingRuleSettings(newRankingRules).getUid());
         String[] updatedRankingRuleSettings = index.getRankingRuleSettings();
 
+        index.waitForTask(index.resetRankingRuleSettings().getUid());
```
brunoocasali, line 204: Why this had to change?

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -210,12 +236,12 @@ public void testUpdateSynonymsSettings() throws Exception {
...
 
         index.waitForTask(index.updateSynonymsSettings(newSynonymsSettings).getUid());
-        Map<String, String[]> updatedRankingRuleSettings = index.getSynonymsSettings();
+        Map<String, String[]> updatedSynonymsSettings = index.getSynonymsSettings();
 
-        assertEquals(newSynonymsSettings.size(), updatedRankingRuleSettings.size());
-        assertEquals(newSynonymsSettings.keySet(), updatedRankingRuleSettings.keySet());
-        assertNotEquals(synonymsSettings.size(), updatedRankingRuleSettings.size());
-        assertNotEquals(synonymsSettings.keySet(), updatedRankingRuleSettings.keySet());
+        assertEquals(newSynonymsSettings.size(), updatedSynonymsSettings.size());
```
brunoocasali, line 241: It seems just a harmful change, but when you are reviewing is hard to understand all the contexts of a PR. In this case, this refactor could be a different PR pointing directly to the `main` ;)

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -476,50 +507,133 @@ public void testUpdateDistinctAttributeSettings() throws Exception {
...
+    @DisplayName("Test get typo tolerance settings by uid")
+    public void testGetTypoTolerance() throws Exception {
+        Index index = createIndex("testGetTypoTolerance");
+        Settings initialSettings = index.getSettings();
+        TypoTolerance initialTypoTolerance = index.getTypoToleranceSettings();
 
-        index.waitForTask(index.updateRankingRuleSettings(newRankingRules).getUid());
-        String[] newRankingRule = index.getRankingRuleSettings();
+        assertEquals(
+                initialSettings.getTypoTolerance().getDisableOnWords().length,
```
brunoocasali, line 536: Why not just assert by the value? It does not work 😬 ?

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -476,50 +507,133 @@ public void testUpdateDistinctAttributeSettings() throws Exception {
...
+                initialSettings.getTypoTolerance().getDisableOnAttributes().length,
+                initialTypoTolerance.getDisableOnAttributes().length);
+        assertEquals(
+                initialSettings.getTypoTolerance().isEnabled(), initialTypoTolerance.isEnabled());
+        assertTrue(
+                initialTypoTolerance.getMinWordSizeForTypos().containsKey("oneTypo")
+                        && initialTypoTolerance.getMinWordSizeForTypos().get("oneTypo") != null);
+        assertTrue(
+                initialTypoTolerance.getMinWordSizeForTypos().containsKey("twoTypos")
+                        && initialTypoTolerance.getMinWordSizeForTypos().get("twoTypos") != null);
```
brunoocasali, line 548: You can use https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertNotNull-java.lang.Object- for this case :)

src/main/java/com/meilisearch/sdk/SettingsHandler.java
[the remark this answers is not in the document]
alallema: Yes, indeed my mistake. I don't think it's a good practice to overload because normally the class should be sent as an argument, no?

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -476,50 +507,133 @@ public void testUpdateDistinctAttributeSettings() throws Exception {
...
+                    }
+                };
+        newTypoTolerance.setMinWordSizeForTypos(minWordSizeTypos);
+        index.waitForTask(index.updateTypoToleranceSettings(newTypoTolerance).getUid());
+        TypoTolerance updatedTypoTolerance = index.getTypoToleranceSettings();
 
-        index.waitForTask(index.updateRankingRuleSettings(null).getUid());
-        String[] resetRankingRule = index.getRankingRuleSettings();
+        assertEquals(
+                newTypoTolerance.getDisableOnWords().length,
```
brunoocasali, line 573: Same thing here: Why not just assert by the value? It does not work 😬 ?

brunoocasali (commented):
This time I could review the tests, awesome work @alallema :D

src/main/java/com/meilisearch/sdk/SettingsHandler.java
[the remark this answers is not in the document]
brunoocasali: I'm not sure, I was just curious about the fact of two different ways to do that, but I think if we can omit the <Task>, we should do it. 
Or if is the case, omit the last parameter (the `Task.class`) and use the `<Task>`, but is up to you in this case, as you wish!

src/main/java/com/meilisearch/sdk/Index.java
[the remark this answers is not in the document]
alallema: I can't I can change it for:
```java
     * @throws Exception if the client has an error
     * @throws MeiliSearchApiException if the response is an error
     */
    public TypoTolerance getTypoToleranceSettings() throws Exception, MeiliSearchApiException {
        return this.settingsHandler.getTypoToleranceSettings(this.uid);
    }
```
Or, I can throw it with a `try` `catch` block. I don't know how you want to do it?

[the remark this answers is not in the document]
brunoocasali: Sorry, I should be more clear about what I asked!

I mean, is good to only `throw` errors that we can control, `Exception` is Ok, but in the end is easier to the consumer just to expect a MeiliSearchApiException instead of the `Exception`, because they will have more control over the exception catch flow like this:

```java
// when we just throw a Meilisearch error, the user can catch all the Meilisearch errors first:
// of course, they need to respect an inheritance (we can talk about this later).

try {
  client.index("movies").anyMethodThatThrowsSomething();
} catch(MeiliSearchOtherException e) {
  // do something with a meilisearch error.
} catch(MeiliSearchActionException e) {
  // do something with a meilisearch error.
} catch(MeiliSearchApiException e) {
  // do something with a meilisearch error.
}
```

Now if we just throw an Exception:

```java
// in this new case, we can see the code is different and harder to read because they had to handle "multiple" 
// different exception possibilities or even check if this exception isn't coming from the okhttp for example.

try {
  client.index("movies").anyMethodThatThrowsSomething();
} catch(Exception e) {
   if (e.message.startsWith("message")) // do something
   else if (e.message.startsWith("other message")) // do other thing 
}
```

src/main/java/com/meilisearch/sdk/TypoTolerance.java
alallema: Like we talk about it it's not the same `Object` it didn't work maybe we can let it that way and open an issue about it? Except if you had another idea to do it

src/test/java/com/meilisearch/integration/SettingsTest.java
alallema: To group the test steps as you suggest, I moved this one line in the `setup` steps so that all `asserts` are grouped afterward in the `verify` steps

src/main/java/com/meilisearch/sdk/TypoTolerance.java
brunoocasali: Yeap, good point!

src/test/java/com/meilisearch/integration/SettingsTest.java
alallema: I can't assert just `getDisableOnWords()` because is a `String[]` I could assert every value in the array but I felt it was too much than just comparing the length of the array don't you think?

.code-samples.meilisearch.yaml
```diff
@@ -192,6 +192,22 @@ update_displayed_attributes_1: |-
   });
 reset_displayed_attributes_1: |-
   client.index("movies").resetDisplayedAttributesSettings();
+get_typo_tolerance_1:
+  client.index("books").getTypoToleranceSettings();
+update_typo_tolerance_1: |-
+  TypoTolerance typoTolerance = new TypoTolerance();
```
brunoocasali, line 198: What about this:

```
TypoTolerance typoTolerance = new TypoTolerance();
HashMap<String, Integer> minWordSizeTypos =
        new HashMap<String, Integer>() {
            {
                put("oneTypo", 4);
                put("twoTypos", 10);
            }
        };

typoTolerance.setMinWordSizeForTypos(minWordSizeTypos);
typoTolerance.setDisableOnAttributes(new String[] {"title"});

client.index("books").updateTypoToleranceSettings(typoTolerance);
```

.code-samples.meilisearch.yaml
```diff
@@ -192,6 +192,22 @@ update_displayed_attributes_1: |-
...
+  typoTolerance.setDisableOnAttributes(new String[] {"title"});
+  HashMap<String, Integer> minWordSizeTypos =
+          new HashMap<String, Integer>() {
+              {
+                  put("oneTypo", 4);
+                  put("twoTypos", 10);
+              }
+          };
+  typoTolerance.setMinWordSizeForTypos(minWordSizeTypos);
+  client.index("movies").updateTypoToleranceSettings(typoTolerance);
```
brunoocasali, line 208: If you disagree is ok, but you at least need to check the index name :)

```suggestion
  client.index("books").updateTypoToleranceSettings(typoTolerance);
```

.code-samples.meilisearch.yaml
```diff
@@ -334,6 +350,40 @@ settings_guide_sortable_1: |-
     "author",
   });
   client.index("books").updateSettings(settings);
+settings_guide_typo_tolerance_1: |-
+  TypoTolerance typoTolerance = new TypoTolerance();
```
brunoocasali, line 354: If you like the other suggestion you can apply here :)

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -476,50 +508,129 @@ public void testUpdateDistinctAttributeSettings() throws Exception {
...
 
-        index.waitForTask(index.updateRankingRuleSettings(null).getUid());
-        String[] resetRankingRule = index.getRankingRuleSettings();
+        assertEquals(
+                newTypoTolerance.getDisableOnWords()[0],
+                updatedTypoTolerance.getDisableOnWords()[0]);
+        assertEquals(
+                newTypoTolerance.getDisableOnAttributes()[0],
+                updatedTypoTolerance.getDisableOnAttributes()[0]);
+        assertEquals(newTypoTolerance.isEnabled(), updatedTypoTolerance.isEnabled());
```
brunoocasali, line 575: maybe just?
```suggestion
        assertTrue(updatedTypoTolerance.isEnabled());
```

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -476,50 +508,129 @@ public void testUpdateDistinctAttributeSettings() throws Exception {
...
+                initialSettings.getTypoTolerance().isEnabled(), initialTypoTolerance.isEnabled());
+        assertNotNull(initialTypoTolerance.getMinWordSizeForTypos().containsKey("oneTypo"));
+        assertNotNull(initialTypoTolerance.getMinWordSizeForTypos().get("oneTypo"));
+        assertNotNull(initialTypoTolerance.getMinWordSizeForTypos().containsKey("twoTypos"));
+        assertNotNull(initialTypoTolerance.getMinWordSizeForTypos().get("twoTypos"));
+    }
+
+    @Test
+    @DisplayName("Test update typo tolerance settings")
+    public void testUpdateTypoTolerance() throws Exception {
```
brunoocasali, line 550: Can you change this test (or add another) with the update of just a couple of fields like only `setDisableOnAttributes` and `minWordSizeTypos`? Just to ensure all the other fields remain the same (because the update is partial)

alallema: I already check this field, you mean check it before the update?

brunoocasali: The assertions phase is great, I mean the exercise part is my point.

Your test is perfect as well, I just would like to have another test that is more "real-life", which means in real life the users may not update all the fields at once like the test is doing, they maybe just change one or two, and my suggestion is to have test with this kind of case.

I don't know if it was clear enough, but let me know!

brunoocasali: btw @alallema this is the last comment, after this I'll approve it 👯

bidoubiwa (approved):
LGTM ! Only issues are not linked to this PR and can be found in the issues section: 

#392
#393 
#394

PR #374, opened 2022-04-25
Ensure nested field support

alallema:
Add tests on nested fields to ensure it is still maintained in Meilisearch.

### DIFF
--- a/src/test/java/com/meilisearch/integration/SearchNestedTest.java
+++ b/src/test/java/com/meilisearch/integration/SearchNestedTest.java
@@ -10,7 +10,6 @@ import com.meilisearch.sdk.Settings;
 import com.meilisearch.sdk.Task;
 import com.meilisearch.sdk.json.GsonJsonHandler;
 import com.meilisearch.sdk.utils.Movie;
-import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


src/test/java/com/meilisearch/integration/SearchNestedTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("integration")
+public class SearchNestedTest extends AbstractIT {
+
+    private TestData<Movie> testData;
+
+    final class Results {
```
brunoocasali, line 23: This class already exist in other places right?

src/test/java/com/meilisearch/integration/SearchNestedTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @BeforeEach
+    public void initialize() {
+        this.setUp();
+        if (testData == null) testData = this.getTestData(NESTED_MOVIES, Movie.class);
+    }
+
+    @AfterAll
+    static void cleanMeiliSearch() {
+        cleanup();
```
brunoocasali, line 41: Hey, look here: https://docs.cypress.io/guides/references/best-practices#Using-after-or-afterEach-hooks

It is not so good to teardown your test setup after the execution :D

src/test/java/com/meilisearch/integration/SearchNestedTest.java
alallema: Yes, in the `SearchTest` file.

src/test/java/com/meilisearch/integration/SearchNestedTest.java
alallema: I don't understand, do you mean I have to do it `afterEach` instead or not clean at all?
Because I like to clean after each test file in order to have independent test no?

brunoocasali (approved):
🍕 awesome @alallema :D

PR #375, opened 2022-04-26
Add new search parameters highlightPreTag, highlightPostTag and cropMarker

alallema:
As per https://github.com/meilisearch/meilisearch/issues/2214 new search parameters are introduced: 

- `highlightPreTag`
- `highlightPostTag`
- `cropMarker`

### DIFF
--- a/src/main/java/com/meilisearch/sdk/Search.java
+++ b/src/main/java/com/meilisearch/sdk/Search.java
@@ -40,7 +40,7 @@ public class Search {
      * @param attributesToRetrieve Attributes to display in the returned documents
      * @param attributesToCrop Attributes whose values have been cropped
      * @param cropLength Length used to crop field values
-     * @param cropMarker String to customize default crop marker, default value: …
+     * @param cropMarker String to add before and/or after the cropped text, default value: …
      * @param highlightPreTag String to customize highlight tag before every highlighted query
      *     terms, default value: <em>
      * @param highlightPostTag String to customize highlight tag after every highlighted query
--- a/src/main/java/com/meilisearch/sdk/Search.java
+++ b/src/main/java/com/meilisearch/sdk/Search.java
@@ -40,6 +40,11 @@ public class Search {
      * @param attributesToRetrieve Attributes to display in the returned documents
      * @param attributesToCrop Attributes whose values have been cropped
      * @param cropLength Length used to crop field values
+     * @param cropMarker String to customize default crop marker, default value: …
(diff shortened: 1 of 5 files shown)


src/main/java/com/meilisearch/sdk/Search.java
```diff
@@ -40,6 +40,11 @@ String rawSearch(String uid, String q) throws Exception {
      * @param attributesToRetrieve Attributes to display in the returned documents
      * @param attributesToCrop Attributes whose values have been cropped
      * @param cropLength Length used to crop field values
+     * @param cropMarker String to customize default crop marker, default value: …
```
brunoocasali, line 43: I think this could be improved to something more informative, without a context the user will not know what is a `crop marker` :)

src/test/java/com/meilisearch/integration/SearchTest.java
```diff
@@ -147,11 +147,37 @@ public void testSearchCrop() throws Exception {
         SearchRequest searchRequest =
                 new SearchRequest("and")
                         .setAttributesToCrop(new String[] {"overview"})
-                        .setCropLength(5);
+                        .setCropLength(1);
 
-        SearchResult searchResult = index.search(searchRequest);
+        Results res_gson = jsonGson.decode(index.rawSearch(searchRequest), Results.class);
```
brunoocasali, line 152: It is not common to define java variables with snake_case :)

src/test/java/com/meilisearch/integration/SearchTest.java
```diff
@@ -147,11 +147,37 @@ public void testSearchCrop() throws Exception {
...
 
-        assertEquals(20, searchResult.getHits().size());
+        assertEquals(20, res_gson.hits.length);
+        assertEquals(5, res_gson.hits[0].getFormatted().getOverview().length());
```
brunoocasali, line 155: `.length()` is a function or a field? Because in the line above you used as a field 🤔 (probably is related to the getOverview response, but just checking out)

src/test/java/com/meilisearch/integration/SearchTest.java
alallema: I know but all the others was this way so I let it the same. I will changes all of them

src/main/java/com/meilisearch/sdk/Search.java
alallema: I modify it, it is better?

brunoocasali (approved):
🥇

PR #427, opened 2022-08-09
Remove duplicate code

### LINKED ISSUE #315: Cleaning and removing duplicate code
The Java client has been completely rewritten by a contributor to use an API based on ServiceTemplate #154.

This refactoring was never implemented because it would have brought too many constraints to the use of this SDK. However, a large part of the code base of this refactor is still present in the SDK in the `src/main/java/com/meilisearch/sdk/api` folder.

The code has been kept since for different reasons:
- the new code base is of very good quality 
- we don't have enough time to consider a refactor and cleanup of the codebase

A refactor of this SDK is cleaning is therefore essential, it makes us lose time during the release because the code is duplicated in some places. It would be necessary to recover the code of the API Service Template and to apply it to the SDK already present which functions while keeping the ease of use.

alallema:
closed #315

A big part of the code of this `SDK` was duplicated and not used, this PR try to remove as much as possible.

**Warning**
This deletion removes some methods that only exist in this part of the code:
- `getStats`
- `getHealth`


brunoocasali (approved):
🗑️

PR #432, opened 2022-08-17
Standardize the `jsonHandler` and reduce the use to one `JSON` library

### LINKED ISSUE #253: Agree on a Json library
I was a bit surprised when I saw that in the project on the one hand Gson and on the other hand org.json.Json is used. I think it would be better to agree on a framework. On the other hand I noticed that the [Settings](https://github.com/meilisearch/meilisearch-java/blob/4d2a6cb83dd58be16d4b29dd1250f990ba0d5fee/src/main/java/com/meilisearch/sdk/Settings.java#L31) contains logic where I think it would be better in `SettingsHandler`. It would be easier to map this logic with Gson. According to Gson the following applies: 

>The default behaviour that is implemented in Gson is that null object fields are ignored. This allows for a more compact output format; however, the client must define a default value for these fields as the JSON format is converted back into its Java form.

Let me know what you think about it

alallema:
# Pull Request

## What does this PR do?
This PR tends to normalize the JSON Library used in this SDK  try to gather all the code using a JSON library into a single handler that can be customizable.

## PR checklist
- [x] fixed #253 (TO FINISH)
- [x] Add the possibility to define a `JSON` through the `Config()` object
- [x] Remove all usage of a Library without passing by the `jsonHandler` `Class`
- [ ] Add a series of tests for the other `jsonHandler` than `GSON`
- [x] Add usage explanation on the README

**Notes**
- Finish removing `org.json.JSON`
- Issue with `getDistinctAttributeSettings()` method who is still messy


src/main/java/com/meilisearch/sdk/Index.java
```diff
@@ -656,7 +646,7 @@ public void fetchPrimaryKey() throws Exception {
         String requestQuery = "/indexes/" + this.uid;
         MeiliSearchHttpRequest meilisearchHttpRequest = new MeiliSearchHttpRequest(config);
         Index retrievedIndex =
-                new Gson().fromJson(meilisearchHttpRequest.get(requestQuery), Index.class);
+                config.jsonHandler.decode(meilisearchHttpRequest.get(requestQuery), Index.class);
```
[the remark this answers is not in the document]
alallema: Yes, I was thinking about that too, but I'll do it when I modify the client I think

src/main/java/com/meilisearch/sdk/Config.java
```diff
@@ -1,20 +1,23 @@
...
+    protected final JsonHandler jsonHandler;
 
     /**
      * Creates a configuration without an API key
      *
      * @param hostUrl URL of the Meilisearch instance
      */
     public Config(String hostUrl) {
-        this(hostUrl, "");
+        this(hostUrl, "", new GsonJsonHandler());
```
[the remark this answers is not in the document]
alallema: > I'm not sure about this constructor. It assumes the apiKey is empty, sounds a little bit reckless.

It does, but `Meilisearch` supports the ability to run without any key, so the SDK can do that too.

> About the idea of calling Gson directly

Yes ! My idea was to have `Gson` by default so we have this one hard written and we advise to use this one but if the user want to use another one he could change it. The problem is that if we don't write a hardcoded choice, the user will have to instantiate one when creating a Client and this will add a step, which is not very recommendable for most users and would add complexity.

src/main/java/com/meilisearch/sdk/model/Result.java
```diff
@@ -1,21 +1,11 @@
 package com.meilisearch.sdk.model;
 
-import com.google.gson.Gson;
 import lombok.Getter;
 
 /** Data structure of MeiliSearch response for a Result */
+@Getter
```
brunoocasali, line 6: Why do we have a `@Getter` here?

src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
```diff
@@ -45,8 +50,7 @@ public <T> T decode(Object o, Class<?> targetClass, Class<?>... parameters) thro
                 return gson.fromJson((String) o, parameterized.getType());
             }
         } catch (Exception e) {
-            // todo: use dedicated exception
-            throw new RuntimeException("Error while deserializing: ", e);
+            throw new MeiliSearchRuntimeException(e);
```
brunoocasali, line 53: I think we can create a new custom class of errors to describe the decoding exception here :)

src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
```diff
@@ -16,25 +19,27 @@ public GsonJsonHandler(Gson gson) {
...
     @Override
     @SuppressWarnings("unchecked")
     public <T> T decode(Object o, Class<?> targetClass, Class<?>... parameters) throws Exception {
         if (o == null) {
-            // todo: use dedicated exception
-            throw new RuntimeException("String to deserialize is null");
+            throw new MeiliSearchRuntimeException();
         }
-        if (targetClass == String.class) {
+        if (o != null && targetClass == String.class) {
```
brunoocasali, line 42: I think this `o != null` will never be true because the JVM will break the execution in the line 40 🍣

src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
```diff
@@ -16,25 +19,27 @@ public GsonJsonHandler(Gson gson) {
...
+        if (o != null && o.getClass() == Key.class) {
+            GsonBuilder builder = new GsonBuilder();
+            this.gson = builder.serializeNulls().setDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'").create();
+        }
         try {
             return gson.toJson(o);
         } catch (Exception e) {
-            // todo: use dedicated exception
-            throw new RuntimeException("Error while serializing: ", e);
+            throw new MeiliSearchRuntimeException(e);
```
brunoocasali, line 32: I think we can create a new custom class of errors to describe the decoding exception here :)

src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
```diff
@@ -16,25 +19,27 @@ public GsonJsonHandler(Gson gson) {
 
     @Override
     public String encode(Object o) throws Exception {
-        if (o.getClass() == String.class) {
+        if (o != null && o.getClass() == String.class) {
             return (String) o;
         }
+        if (o != null && o.getClass() == Key.class) {
```
brunoocasali, line 25: Can you put a `TODO: review later` here on this line?

We cannot create those levels of specialization inside of this class because this is an infinite job if other classes start to have similar issues to be handled.

src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -70,15 +73,13 @@ public Task getTask(int taskUid) throws Exception, MeiliSearchApiException {
...
         String urlPath = "/tasks";
 
         Result<Task> result =
-                this.gson.fromJson(
-                        this.meilisearchHttpRequest.get(urlPath),
-                        new TypeToken<Result<Task>>() {}.getType());
-
-        return result.getResults();
+                meilisearchHttpRequest.jsonHandler.decode(
+                        this.meilisearchHttpRequest.get(urlPath), Result.class, Task.class);
```
brunoocasali, line 81: `Result<Task>.class` did not work?

src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -47,9 +45,13 @@ public Task getTask(String indexUid, int taskUid) throws Exception {
...
      * @throws Exception if client request causes an error
      */
-    public Task[] getTasks(String indexUid) throws Exception {
+    public Result<Task> getTasks(String indexUid) throws Exception {
         String urlPath = "/indexes/" + indexUid + "/tasks";
-        return this.gson.fromJson(this.meilisearchHttpRequest.get(urlPath), Task[].class);
+
+        Result<Task> result =
+                meilisearchHttpRequest.jsonHandler.decode(
+                        this.meilisearchHttpRequest.get(urlPath), Result.class, Task.class);
```
brunoocasali, line 53: `Result<Task>.class` did not work?

src/main/java/com/meilisearch/sdk/SearchRequest.java
```diff
@@ -1,10 +1,11 @@
 package com.meilisearch.sdk;
 
 import lombok.Getter;
-import org.json.JSONObject;
+import lombok.Setter;
 
 /** Search request query string builder */
 @Getter
+@Setter
```
brunoocasali, line 8: Are you kidding that `@Setter` config can make a whole builder just by adding the config?

That's why you could remove the `setOffset`, `setQ` and others?

brunoocasali: By the way, what do you think to have a `setQuery` instead of `setQ`?

src/main/java/com/meilisearch/sdk/IndexesHandler.java
```diff
@@ -37,12 +37,12 @@ String create(String uid) throws Exception, MeiliSearchApiException {
      * @throws Exception if an error occurs
      */
     String create(String uid, String primaryKey) throws Exception, MeiliSearchApiException {
-        JsonObject params = new JsonObject();
-        params.addProperty("uid", uid);
-        if (primaryKey != null) {
-            params.addProperty("primaryKey", primaryKey);
-        }
-        return meilisearchHttpRequest.post("/indexes", params.toString());
+        HashMap<String, Object> index = new HashMap<String, Object>();
```
brunoocasali, line 40: What about creating a type to make this easier and reducing the duplication in the updatePK method below?

src/test/java/com/meilisearch/integration/KeysTest.java
```diff
@@ -28,7 +29,8 @@ static void cleanMeiliSearch() {
     /** Test Get Keys */
     @Test
     public void testClientGetKeys() throws Exception {
-        Key[] keys = client.getKeys();
+        Result<Key> result = client.getKeys();
```
brunoocasali, line 32: It shouldn't be `Result<Key[]>` ?

src/test/java/com/meilisearch/integration/SearchTest.java
```diff
@@ -72,14 +71,14 @@ public void testBasicSearch() throws Exception {
...
         String indexUid = "SearchOffset";
         Index index = client.index(indexUid);
-        GsonJsonHandler jsonGson = new GsonJsonHandler();
 
         TestData<Movie> testData = this.getTestData(MOVIES_INDEX, Movie.class);
         Task task = index.addDocuments(testData.getRaw());
 
         index.waitForTask(task.getUid());
 
-        SearchRequest searchRequest = new SearchRequest("a").setOffset(20);
```
brunoocasali, line 82: This way do not work anymore? because of the `@setter`?

brunoocasali: If not you can use this option:
image
https://projectlombok.org/features/GetterSetter and make them a builder again! 🎉

src/test/java/com/meilisearch/integration/TasksTest.java
```diff
@@ -51,7 +52,8 @@ public void testClientGetTask() throws Exception {
     /** Test Get Tasks */
     @Test
     public void testClientGetTasks() throws Exception {
-        Task[] tasks = client.getTasks();
+        Result<Task> result = client.getTasks();
```
brunoocasali, line 55: Shouldn't be `Result<Task[]>`

src/test/java/com/meilisearch/integration/classes/AbstractIT.java
```diff
@@ -80,7 +81,8 @@ public void loadResource(String fileName) throws IOException {
     }
 
     public Key getPrivateKey() throws Exception {
-        Key[] keys = client.getKeys();
+        Result<Key> result = client.getKeys();
+        Key[] keys = result.getResults();
```
brunoocasali, line 85: Same thing here about the different type `Result<Keys[]>`

src/test/java/com/meilisearch/integration/SettingsTest.java
```diff
@@ -125,6 +123,50 @@ public void testUpdateSettingsTypoTolerance() throws Exception {
...
+        Settings settingsDisplayedAttr = new Settings();
+        settingsDisplayedAttr.setDisplayedAttributes(
+                new String[] {"title", "overview", "genres", "release_date"});
+        index.waitForTask(index.updateSettings(settingsDisplayedAttr).getUid());
+        Settings newSettingsDisplayedAttr = index.getSettings();
+
+        Settings settingsRankingRules = new Settings();
+        settingsRankingRules.setRankingRules(
+                new String[] {
+                    "typo",
```
brunoocasali, line 139: Can you just have one or two? Because I'm not sure if the Meilisearch defaults are the same you choose here to test.

src/main/java/com/meilisearch/sdk/SettingsHandler.java
```diff
@@ -90,10 +86,12 @@ public String[] getRankingRuleSettings(String uid) throws Exception {
...
      */
     public Task updateRankingRuleSettings(String uid, String[] rankingRules) throws Exception {
-        String rankingRulesAsJson = gson.toJson(rankingRules);
-        return this.gson.fromJson(
+        return meilisearchHttpRequest.jsonHandler.decode(
                 meilisearchHttpRequest.post(
-                        "/indexes/" + uid + "/settings/ranking-rules", rankingRulesAsJson),
+                        "/indexes/" + uid + "/settings/ranking-rules",
+                        rankingRules == null
+                                ? meilisearchHttpRequest.jsonHandler.encode(rankingRules)
```
brunoocasali, line 93: If you just encode null, does not work?

src/main/java/com/meilisearch/sdk/SearchRequest.java
alallema: > Are you kidding that @Setter config can make a whole builder just by adding the config?

This is cool, right? Actually the `setQuery()` was kind of specific but we don't have the usage of it and Yes I think we should rename `setQ` to `setQuery`

src/test/java/com/meilisearch/integration/KeysTest.java
alallema: No, because it is already a `<T> []` so it just needs the type of with what we will do the `[]`

src/main/java/com/meilisearch/sdk/SettingsHandler.java
alallema: Yes, it works but I find it dirty no?
Actually I can write:
```
                        rankingRules == null
                                ? meilisearchHttpRequest.jsonHandler.encode(null)
                                : rankingRules),
```
And even:
```
                        rankingRules == null ? "null" : rankingRules),
```
What do you think is the best?

src/main/java/com/meilisearch/sdk/model/Result.java
alallema: To be able to get the `Result` who should be `Results` by the way ...
```java
    public void testClientGetKeys() throws Exception {
        Result<Key> result = client.getKeys();
        Key[] keys = result.getResults();
```

src/test/java/com/meilisearch/integration/KeysTest.java
brunoocasali: I see, I think it is easier to understand `Result<Key[]>` or `Results<Key>` or even `KeysResults` (because of the plural). 

This way I tend to believe I will deal with a only one object.

src/main/java/com/meilisearch/sdk/SettingsHandler.java
brunoocasali: > Yes, it works but I find it dirty no?

I don't think so if the idea is to send an empty array of strings the result should be `[]` 🤔

src/test/java/com/meilisearch/integration/SettingsTest.java
alallema: `Meilisearch` has 6 by default:
```
        assertEquals(4, newSettingsDisplayedAttr.getDisplayedAttributes().length);
        assertEquals(6, newSettingsDisplayedAttr.getRankingRules().length);
        assertEquals(0, newSettingsDisplayedAttr.getSynonyms().size());
        assertEquals(4, newSettingsRankingRules.getDisplayedAttributes().length);
        assertEquals(8, newSettingsRankingRules.getRankingRules().length);
        assertEquals(0, newSettingsRankingRules.getSynonyms().size());
        assertEquals(4, newSettingsSynonyms.getDisplayedAttributes().length);
        assertEquals(2, newSettingsSynonyms.getSynonyms().size());
        assertEquals(8, newSettingsSynonyms.getRankingRules().length);
```
That why we check 6 then 8

src/test/java/com/meilisearch/integration/KeysTest.java
alallema: Yes I had to change all `Result` by `Results`

src/main/java/com/meilisearch/sdk/SettingsHandler.java
alallema: The idea is that when you send "null", you can reset the ranking rules setting so that it returns to its default state:
```
[
	"words",
	"typo",
	"proximity",
	"attribute",
	"sort",
	"exactness"
]
```
If we send an empty array it will become:
```
[]
```

src/main/java/com/meilisearch/sdk/IndexesHandler.java
alallema: Yes, I also thought about it but we already have an index object. So I should create a `partialIndex` object or something and I wasn't sure about it ... No?

brunoocasali (approved):

PR #439, opened 2022-09-14
Renaming last  MeiliSearch to Meilisearch

alallema:
# Pull Request

## What does this PR do?
Replace the name`Meilisearch` with the correct typo where it remained.

**Notes**:
`MeiliSearchHttpRequest` will be change in the same time as the Client.


brunoocasali (approved):
🎉 🎉 ✍️

PR #448, opened 2022-09-28
Move HttpClient to Config 

alallema:
The `MeilisearchHttpClient` variable has been moved to the `Config` class in a way to be instantiated once instead of being created in each Handler.


alallema (commented):
Thank you for wanting to contribute to this repository :heart:.
Unfortunately, this repository is not open to contributions from Hacktoberfest. Indeed it is being reworked and will not participate this year.

*PS: This message was sent automatically!*

brunoocasali (approved):

PR #470, opened 2022-10-25
Provide a generic to http delete method

alallema:
# Pull Request

## What does this PR do?
Most of the methods had to use the `jsonHandler` after the call to the `delete` method through the client. The purpose of this PR is to bypass this step to simplify the call to the `delete` method via the client:
```java
    public Task resetRankingRulesSettings(String uid) throws MeilisearchException {
        Task task = jsonHandler.decode(this.indexesHandler.delete(uid), Task.class);
        return task;
    }
```
became
```java
    public Task resetRankingRulesSettings(String uid) throws MeilisearchException {
        String urlPath = "/indexes/" + uid + "/settings/ranking-rules";
        return httpClient.delete(urlPath, Task.class);
    }
```

### DIFF
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -147,8 +147,7 @@ public class Client {
      * @throws MeilisearchException if an error occurs
      */
     public Task deleteIndex(String uid) throws MeilisearchException {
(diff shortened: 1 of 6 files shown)


alallema:
> I'll let you fix the CI before review CC: @alallema

Done, there was an extra curly brace ...

PR #489, opened 2022-11-10
Add testing against different Java versions in CI

alallema:


brunoocasali:
@alallema I think the gradle version does not support Java 17, that's why your CI is breaking!

alallema:
This PR was just a test.
There is no need to add different version matrix on the CI

PR #502, opened 2022-12-12
Developer experience enhancements

brunoocasali:
- Set the indent_style to "space" instead of "tabs".
- Enable test reports to be forwarded to Docker volume
- Allow debugging output from env var.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,7 @@ max_line_length = 120
 
 [*.java]
 max_line_length = 150
-indent_style = tab
+indent_style = space
 indent_size = 4
 tab_width = 4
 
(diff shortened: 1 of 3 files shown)


brunoocasali:
About the space/tabs thing, I just used it because the project is already using that, maybe you want tabs, if true let's move everything in the repo to tabs!

alallema (approved):
Thanks for this PR @brunoocasali ❤️ 
LGTM!
>About the space/tabs thing, I just used it because the project is already using that, maybe you want tabs, if true let's move everything in the repo to tabs!
I don't care actually 😄 I don't even know which one is used my IDE do it for me 😉

alallema:
bors merge

PR #509, opened 2022-12-20
Add URLBuilder for query parameters

alallema:
The v0.28.0 version has added the possibility to filter the results via query parameters.
A new Class `URLBuilder` has been added in this PR to help create URLs with query parameters

PR #510, opened 2022-12-21
Apply tasks changes

alallema:
# Pull Request

## What does this PR do?

### Changes

- Remove `GET /indexes/:indexUid/tasks`. Use `GET /tasks?indexUid=:indexUid` instead.
- Remove `GET /indexes/:indexUid/tasks/:taskUid`. Use `GET /tasks/:taskUid` instead.
- Possibility to filter the results:
  - In the endpoint `GET /tasks` we can filter by: `type`, `status` and `indexUid`.
  - We need to join the array values when filter by `status` like this: `enqueued,processing`.
- [x] `index.getTask` now use `GET /tasks/:task_uid`
- [x] `index.getTasks` now use `GET/tasks?indexUid=movies`
- [x] Rename task `uid` field to `taskUid`.
- [x] `index.getTasks` accept pagination metadata, added limit (default: 20), from.
- [x] `client.getTasks` accept pagination metadata, added limit (default: 20), from.
- [x] Add `TaskInfo` class to handle all response from asynchronous operation
- [x]  Create `TasksQuery` class to handle pagination filtering
- [x]  Create `TasksResults` class to retrieve responses from `getTasks`


src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -26,53 +28,93 @@ public class TasksHandler {
...
+        urlb.addSubroute("tasks");
+        urlb.addParameter("limit", param.getLimit());
+        urlb.addParameter("from", param.getFrom());
+        urlb.addParameter("status", param.getStatus());
+        urlb.addParameter("type", param.getType());
+        urlb.addParameter("indexUid", param.getIndexUid());
+        String urlQuery = urlb.getURL();
+
+        TasksResults result = httpClient.get(urlQuery, TasksResults.class);
+        return result;
```
brunoocasali, line 73: My java can be a little rusty, but I think the code doesn't need to be that much expressive all the time. In this case from my understanding, the variables `result` and `urlQuery` and the multiple `.addParameter` could be changed by something like this:

```suggestion
    URLBuilder urlb = new URLBuilder("tasks")
        .addParameter("limit", param.getLimit())
        .addParameter("from", param.getFrom())
        .addParameter("status", param.getStatus())
        .addParameter("type", param.getType())
        .addParameter("indexUid", param.getIndexUid());

    return httpClient.get(urlb.getURL(), TasksResults.class);
```

The code will not work without my suggestions here https://github.com/meilisearch/meilisearch-java/pull/509#pullrequestreview-1224949920

src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -26,53 +28,93 @@ public class TasksHandler {
...
-     * Retrieves tasks from the client
+     * Retrieves all TasksHandler at the specified index uid
      *
+     * @param indexUid Index identifier to the requested Tasks
      * @return List of task instance
      * @throws MeilisearchException if client request causes an error
      */
-    Results<Task> getTasks() throws MeilisearchException {
-        String urlPath = "/tasks";
+    TasksResults getTasks(String indexUid) throws MeilisearchException {
```
brunoocasali, line 83: The comment above can be applied in the other methods as well.

src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -26,53 +28,93 @@ public class TasksHandler {
...
+        urlb.addParameter("indexUid", param.getIndexUid());
+        String urlQuery = urlb.getURL();
+
+        TasksResults result = httpClient.get(urlQuery, TasksResults.class);
+        return result;
     }
 
     /**
-     * Retrieves tasks from the client
+     * Retrieves all TasksHandler at the specified index uid
```
brunoocasali, line 77: Not sure if I got this... 👀

src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -26,53 +28,93 @@ public class TasksHandler {
...
+        urlb.addParameter("indexUid", indexUid);
+        String urlQuery = urlb.getURL();
+
+        TasksResults result = httpClient.get(urlQuery, TasksResults.class);
+        return result;
+    }
+
+    /**
+     * Retrieves all TasksHandler at the specified index uid
+     *
```
brunoocasali, line 95: Here as well.

src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -26,53 +28,93 @@ public class TasksHandler {
...
+     * @param indexUid Index identifier to the requested Tasks
+     * @param param accept by the tasks route
+     * @return List of task instance
+     * @throws MeilisearchException if client request causes an error
+     */
+    TasksResults getTasks(String indexUid, TasksQuery param) throws MeilisearchException {
+        String[] newIndexUid = new String[param.getIndexUid().length + 1];
```
brunoocasali, line 102: I'm not sure why you had to do this. Can you explain it to me, please?

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -106,15 +106,16 @@ String getDocuments(String uid, int limit, int offset, List<String> attributesTo
...
      * @param document String containing the document to add
      * @param primaryKey PrimaryKey of the document
-     * @return Meilisearch's Task API response
+     * @return Meilisearch's TaskInfo API response
      * @throws MeilisearchException if the client request causes an error
      */
-    Task addDocuments(String uid, String document, String primaryKey) throws MeilisearchException {
+    TaskInfo addDocuments(String uid, String document, String primaryKey)
+            throws MeilisearchException {
         String urlQuery = "/indexes/" + uid + "/documents";
```
brunoocasali, line 114: You can apply the URLBuilder here :)

src/test/java/com/meilisearch/integration/ClientTest.java
```diff
@@ -9,7 +9,8 @@
 import com.meilisearch.integration.classes.TestData;
 import com.meilisearch.sdk.Index;
 import com.meilisearch.sdk.exceptions.MeilisearchApiException;
-import com.meilisearch.sdk.model.Task;
+import com.meilisearch.sdk.model.Results;
```
brunoocasali, line 12: I find no usage for this class here :)

```suggestion
```

src/main/java/com/meilisearch/sdk/Documents.java
alallema: Yes, but I would like to create a specific PR to modify each method with the URLBuilder. WDYT?

src/test/java/com/meilisearch/integration/ClientTest.java
alallema: Not yet ... I remove it

src/main/java/com/meilisearch/sdk/TasksHandler.java
alallema: I modify it it's better?

src/main/java/com/meilisearch/sdk/TasksHandler.java
alallema: It's for this specific case:
```java
    TasksQuery query = new TasksQuery().setIndexes(new String[] {"books", "cities"});
    client.index("movies").getTasks();
```
In case the `getTasks()` method is called through the `IndexesHandler` the request of this should be:
```
/tasks?indexUids=movies,books,cities
```

src/main/java/com/meilisearch/sdk/Documents.java
alallema: Yes this one will be changed during PR about documents changes

src/main/java/com/meilisearch/sdk/TasksHandler.java
brunoocasali: What I found strange was the `TasksHandler` part. Maybe you want to `Retrieves all tasks at the specified index uid`

brunoocasali (approved):

PR #516, opened 2022-12-27
Update Index API for Meilisearch v.28

alallema:
# Pull Request

Related to: https://github.com/meilisearch/meilisearch/issues/2373

### Changes

- [x] `client.getIndexes()` accept pagination metadata, added limit (default: 20), offset (default: 0), total
- [x] Wrap the indexes objects in a `Results` object
- [x] Remove `client.getRawIndex()` method 
- [x] `client.getRawIndexes()` method return a `String`
- [x] Creation of `IndexesQuery` class
- [x] Rename the methods of the "IndexesHandler" to match the name of the Client methods 

```java
    client.getIndexes()
```

returns for example
```bash
com.meilisearch.sdk.model.Results@2a742aa2
```

```java
    client.getRawIndexes()
```

returns for example

```bash
{"results":[],"offset":0,"limit":20,"total":0}
```

### DIFF
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -68,8 +68,7 @@ public class Client {
     }
 
     /**
-     * Gets all indexes
-     * https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
+     * Gets all indexes https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
      *
      * @return Results containing a list of indexes from the Meilisearch API
      * @throws MeilisearchException if an error occurs
@@ -83,10 +82,9 @@ public class Client {
     }
 
     /**
-     * Gets indexes
-     * https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
+     * Gets indexes https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
      *
-     * @param query parameters accepted by the get indexes route
+     * @param params query parameters accepted by the get indexes route
      * @return Results containing a list of indexes from the Meilisearch API
      * @throws MeilisearchException if an error occurs
      */
@@ -99,8 +97,7 @@ public class Client {
     }
 
     /**
-     * Gets all indexes
-     * https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
+     * Gets all indexes https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
      *
      * @return List of indexes from the Meilisearch API as String
      * @throws MeilisearchException if an error occurs
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -7,6 +7,7 @@ import com.auth0.jwt.JWT;
 import com.auth0.jwt.algorithms.Algorithm;
 import com.meilisearch.sdk.exceptions.MeilisearchException;
 import com.meilisearch.sdk.json.JsonHandler;
+import com.meilisearch.sdk.model.IndexesQuery;
 import com.meilisearch.sdk.model.Key;
 import com.meilisearch.sdk.model.Results;
 import com.meilisearch.sdk.model.Stats;
@@ -63,7 +64,7 @@ public class Client {
      * @throws MeilisearchException if an error occurs
      */
     public TaskInfo createIndex(String uid, String primaryKey) throws MeilisearchException {
-        return this.indexesHandler.create(uid, primaryKey);
+        return this.indexesHandler.createIndex(uid, primaryKey);
     }
 
     /**
@@ -73,9 +74,25 @@ public class Client {
      * @return Array of indexes in the Meilisearch client
      * @throws MeilisearchException if an error occurs
      */
-    public Index[] getIndexes() throws MeilisearchException {
-        Index[] indexes = jsonHandler.decode(getRawIndexes(), Index[].class);
-        for (Index index : indexes) {
+    public Results<Index> getIndexes() throws MeilisearchException {
+        Results<Index> indexes = this.indexesHandler.getIndexes();
+        for (Index index : indexes.getResults()) {
+            index.setConfig(this.config);
+        }
+        return indexes;
+    }
+
+    /**
+     * Gets indexes in the current Meilisearch instance
+     * https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
+     *
+     * @param param accept by the indexes route
+     * @return Array of indexes in the Meilisearch client
+     * @throws MeilisearchException if an error occurs
+     */
+    public Results<Index> getIndexes(IndexesQuery param) throws MeilisearchException {
+        Results<Index> indexes = this.indexesHandler.getIndexes(param);
+        for (Index index : indexes.getResults()) {
             index.setConfig(this.config);
         }
         return indexes;
@@ -89,7 +106,7 @@ public class Client {
      * @throws MeilisearchException if an error occurs
      */
     public String getRawIndexes() throws MeilisearchException {
-        return this.indexesHandler.getAll();
+        return this.indexesHandler.getRawIndexes();
     }
 
     /**
@@ -117,21 +134,9 @@ public class Client {
      * @throws MeilisearchException if an error occurs
      */
     public Index getIndex(String uid) throws MeilisearchException {
-        Index index = jsonHandler.decode(getRawIndex(uid), Index.class);
+        Index index = this.indexesHandler.getIndex(uid);
         index.setConfig(this.config);
         return index;
-    }
-
-    /**
-     * Gets single index by its unique identifier
-     * https://docs.meilisearch.com/reference/api/indexes.html#get-one-index
-     *
-     * @param uid Unique identifier of the index to get
-     * @return Meilisearch API response as String
-     * @throws MeilisearchException if an error occurs
-     */
-    public String getRawIndex(String uid) throws MeilisearchException {
-        return this.indexesHandler.get(uid);
     }
 
     /**
@@ -156,7 +161,7 @@ public class Client {
      * @throws MeilisearchException if an error occurs
      */
     public TaskInfo deleteIndex(String uid) throws MeilisearchException {
-        return this.indexesHandler.delete(uid);
+        return this.indexesHandler.deleteIndex(uid);
     }
 
     /**
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -90,8 +90,8 @@ public class Client {
      * @return Results containing a list of indexes from the Meilisearch API
      * @throws MeilisearchException if an error occurs
      */
-    public Results<Index> getIndexes(IndexesQuery param) throws MeilisearchException {
-        Results<Index> indexes = this.indexesHandler.getIndexes(param);
+    public Results<Index> getIndexes(IndexesQuery params) throws MeilisearchException {
+        Results<Index> indexes = this.indexesHandler.getIndexes(params);
         for (Index index : indexes.getResults()) {
             index.setConfig(this.config);
         }
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -268,7 +268,7 @@ public class Client {
      * https://docs.meilisearch.com/reference/api/keys.html#get-one-key
      *
      * @param uid Identifier of the requested Key
-     * @return Meilisearch API response as Key Instance
+     * @return Key Instance
      * @throws MeilisearchException if an error occurs
      */
     public Key getKey(String uid) throws MeilisearchException {
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -68,7 +68,7 @@ public class Client {
     }
 
     /**
-     * Gets all indexes https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
+     * Gets indexes https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
      *
      * @return Results containing a list of indexes from the Meilisearch API
      * @throws MeilisearchException if an error occurs
--- a/src/main/java/com/meilisearch/sdk/Client.java
+++ b/src/main/java/com/meilisearch/sdk/Client.java
@@ -68,7 +68,7 @@ public class Client {
     }
 
     /**
-     * Gets all indexes in the current Meilisearch instance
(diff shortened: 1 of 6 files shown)


src/main/java/com/meilisearch/sdk/IndexesHandler.java
```diff
@@ -67,7 +72,33 @@ String get(String uid) throws MeilisearchException {
...
+    /**
+     * Gets indexes in the current Meilisearch instance
+     *
+     * @param param accept by the indexes route
+     * @return Meilisearch API response
+     * @throws MeilisearchException if an error occurs
+     */
+    Results<Index> getIndexes(IndexesQuery param) throws MeilisearchException {
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("indexes")
```
brunoocasali, line 88: I pointed in the other PRs a similar approach where we introduce this code inside of the query handler :)

src/test/java/com/meilisearch/integration/classes/AbstractIT.java
```diff
@@ -105,7 +105,8 @@ public static void deleteAllIndexes() {
             Client ms = new Client(new Config(getMeilisearchHost(), "masterKey"));
             Results<Index> indexes = ms.getIndexes();
             for (Index index : indexes.getResults()) {
-                ms.deleteIndex(index.getUid());
+                TaskInfo task = ms.deleteIndex(index.getUid());
```
brunoocasali, line 108: Nice change! 👯

brunoocasali (changes requested):

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -73,9 +74,25 @@ public TaskInfo createIndex(String uid, String primaryKey) throws MeilisearchExc
      * @return Array of indexes in the Meilisearch client
```
bidoubiwa, line 74: It's not an array anymore but a Results

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -73,9 +74,25 @@ public TaskInfo createIndex(String uid, String primaryKey) throws MeilisearchExc
...
+        }
+        return indexes;
+    }
+
+    /**
+     * Gets indexes in the current Meilisearch instance
+     * https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
+     *
+     * @param param accept by the indexes route
+     * @return Array of indexes in the Meilisearch client
```
bidoubiwa, line 90: not an array of indexes but a results

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -73,9 +74,25 @@ public TaskInfo createIndex(String uid, String primaryKey) throws MeilisearchExc
...
+            index.setConfig(this.config);
+        }
+        return indexes;
+    }
+
+    /**
+     * Gets indexes in the current Meilisearch instance
+     * https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
+     *
+     * @param param accept by the indexes route
```
bidoubiwa, line 89:
```suggestion
     * @param query parameters accepted by the get indexes route
```

bidoubiwa (changes requested):

bidoubiwa (commented):
I don't see the `updateIndex` method.

Can you also change the name of the PR with more context? For example

> Update Index API for Meilisearch v.28

src/main/java/com/meilisearch/sdk/IndexesHandler.java
```diff
@@ -56,9 +59,11 @@ TaskInfo create(String uid, String primaryKey) throws MeilisearchException {
...
      * @throws MeilisearchException if an error occurs
      */
-    String get(String uid) throws MeilisearchException {
-        String requestQuery = "/indexes/" + uid;
-        return httpClient.get(requestQuery, String.class);
+    Index getIndex(String uid) throws MeilisearchException {
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("indexes").addSubroute(uid);
+        String urlPath = urlb.getURL();
+        return httpClient.get(urlPath, Index.class);
```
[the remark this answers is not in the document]
alallema: Of course, it still works, but I create a new `URLBuilder` class, so I used it everywhere as Bruno suggest it.

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -89,7 +106,7 @@ public Index[] getIndexes() throws MeilisearchException {
      * @throws MeilisearchException if an error occurs
      */
     public String getRawIndexes() throws MeilisearchException {
```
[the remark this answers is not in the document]
alallema: Not yet, but I will create an issue if someone want to implement it 😃

src/main/java/com/meilisearch/sdk/IndexesHandler.java
```diff
@@ -67,7 +72,28 @@ String get(String uid) throws MeilisearchException {
...
+        return httpClient.get(param.toQuery(param), Results.class, Index.class);
+    }
+
+    /**
+     * Gets all indexes in the current Meilisearch instance
+     *
+     * @return Meilisearch API response
+     * @throws MeilisearchException if an error occurs
+     */
+    String getRawIndexes() throws MeilisearchException {
```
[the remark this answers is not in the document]
alallema: Same as below

src/main/java/com/meilisearch/sdk/model/IndexesQuery.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * <p>https://docs.meilisearch.com/reference/api/indexes.html#query-parameters
+ */
+@Setter
+@Getter
+@Accessors(chain = true)
+public class IndexesQuery {
+    private int offset = -1;
+    private int limit = -1;
+
```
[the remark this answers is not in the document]
alallema: Yes and no. It's not really clean but this default value allows us to check that these values have been set by the user and in this case to encode them as a query parameters

alallema:
> I don't see the `updateIndex` method.
> 
> Can you also change the name of the PR with more context? For example
> 
> > Update Index API for Meilisearch v.28

It's on line 151 on `Client` file and this method is called `updatePrimaryKey` on the `IndexesHandler` file because it updates the primary key.

src/main/java/com/meilisearch/sdk/Client.java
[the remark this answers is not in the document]
alallema: #521 🎉

src/main/java/com/meilisearch/sdk/IndexesHandler.java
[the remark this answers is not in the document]
bidoubiwa: Should it work the same way in the `Keys` PR ? #517

src/main/java/com/meilisearch/sdk/IndexesHandler.java
```diff
@@ -53,21 +56,44 @@ TaskInfo create(String uid, String primaryKey) throws MeilisearchException {
...
+
+    /**
+     * Gets indexes in the current Meilisearch instance
+     *
+     * @param query parameters accepted by the indexes route
+     * @return Results containing a list of indexes
+     * @throws MeilisearchException if an error occurs
+     */
+    Results<Index> getIndexes(IndexesQuery param) throws MeilisearchException {
+        return httpClient.get(param.toQuery(param), Results.class, Index.class);
```
bidoubiwa, line 87:
```suggestion
    Results<Index> getIndexes(IndexesQuery params) throws MeilisearchException {
        return httpClient.get(param.toQuery(params), Results.class, Index.class);
```

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -63,33 +64,49 @@ public TaskInfo createIndex(String uid) throws MeilisearchException {
...
+     * @return Results containing a list of indexes from the Meilisearch API
+     * @throws MeilisearchException if an error occurs
+     */
+    public Results<Index> getIndexes(IndexesQuery param) throws MeilisearchException {
+        Results<Index> indexes = this.indexesHandler.getIndexes(param);
+        for (Index index : indexes.getResults()) {
+            index.setConfig(this.config);
+        }
+        return indexes;
+    }
```
bidoubiwa, line 99:
```suggestion
    public Results<Index> getIndexes(IndexesQuery params) throws MeilisearchException {
        Results<Index> indexes = this.indexesHandler.getIndexes(params);
        for (Index index : indexes.getResults()) {
            index.setConfig(this.config);
        }
        return indexes;
    }
```

src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -58,7 +58,7 @@ TasksResults getTasks() throws MeilisearchException {
      * Retrieves all tasks from the client
      *
      * @param param accept by the tasks route
-     * @return List of task instance
+     * @return TasksResults containing a list of task instance
```
bidoubiwa, line 61: All the @returns changes on non index related methods should not be in this PR but if you want 🙈

src/test/java/com/meilisearch/integration/ClientTest.java
```diff
@@ -28,10 +30,6 @@ public void initialize() {
         setUp();
         setUpJacksonClient();
         if (testData == null) testData = this.getTestData(MOVIES_INDEX, Movie.class);
-    }
-
-    @AfterAll
-    static void cleanMeilisearch() {
         cleanup();
     }
```
bidoubiwa, line 34: I think you can also keep the `afterAll` otherwise the last index of the last test will not be deleted

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -263,7 +268,7 @@ public void waitForTask(int uid) throws MeilisearchException {
      * https://docs.meilisearch.com/reference/api/keys.html#get-one-key
      *
      * @param uid Identifier of the requested Key
-     * @return Key Instance
+     * @return Meilisearch API response as Key Instance
```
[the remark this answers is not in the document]
alallema: This one can be on the Key changes 😃

src/main/java/com/meilisearch/sdk/Client.java
```diff
@@ -63,33 +64,46 @@ public TaskInfo createIndex(String uid) throws MeilisearchException {
...
      */
     public TaskInfo createIndex(String uid, String primaryKey) throws MeilisearchException {
-        return this.indexesHandler.create(uid, primaryKey);
+        return this.indexesHandler.createIndex(uid, primaryKey);
     }
 
     /**
-     * Gets all indexes in the current Meilisearch instance
-     * https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
+     * Gets all indexes https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
```
bidoubiwa, line 71:
```suggestion
     * Gets indexes https://docs.meilisearch.com/reference/api/indexes.html#list-all-indexes
```

src/main/java/com/meilisearch/sdk/IndexesHandler.java
```diff
@@ -53,21 +55,41 @@ TaskInfo create(String uid, String primaryKey) throws MeilisearchException {
...
      */
-    String get(String uid) throws MeilisearchException {
-        String requestQuery = "/indexes/" + uid;
-        return httpClient.get(requestQuery, String.class);
+    Index getIndex(String uid) throws MeilisearchException {
+        return httpClient.get(new IndexesQuery().toQuery(uid), Index.class);
     }
 
     /**
      * Gets all indexes in the current Meilisearch instance
```
bidoubiwa, line 66:
```suggestion
     * Gets indexes in the current Meilisearch instance
```

I removed the `all` because they are now returned paginated thus they might not all be returned

src/main/java/com/meilisearch/sdk/IndexesHandler.java
```diff
@@ -53,21 +55,41 @@ TaskInfo create(String uid, String primaryKey) throws MeilisearchException {
...
      *
-     * @return Meilisearch API response
+     * @return Results containing a list of indexes
      * @throws MeilisearchException if an error occurs
      */
-    String getAll() throws MeilisearchException {
+    Results<Index> getIndexes() throws MeilisearchException {
+        return httpClient.get("/indexes", Results.class, Index.class);
```
bidoubiwa, line 72: It could be nice to use `IndexesQuery` everywhere so that we only hardcode `/indexes` in `toQuery`. Nonetheless, maybe it is worth creating an issue for that change to avoid making this PR more complex

src/main/java/com/meilisearch/sdk/IndexesHandler.java
```diff
@@ -53,21 +55,41 @@ TaskInfo create(String uid, String primaryKey) throws MeilisearchException {
...
+     * @param query parameters accepted by the indexes route
+     * @return Results containing a list of indexes
+     * @throws MeilisearchException if an error occurs
+     */
+    Results<Index> getIndexes(IndexesQuery params) throws MeilisearchException {
+        return httpClient.get(params.toQuery(params), Results.class, Index.class);
+    }
+
+    /**
+     * Gets all indexes in the current Meilisearch instance
```
bidoubiwa, line 87:
```suggestion
     * Gets indexes in the current Meilisearch instance
```

bidoubiwa (changes requested):

src/main/java/com/meilisearch/sdk/IndexesHandler.java
alallema: We can indeed do this but it means creating a new function that will require resources and complicate the general call such as:
`return httpClient.get("/indexes", Results.class, Index.class);`
will became:
`return httpClient.get(new IndexesQuery().toQuery(), Results.class, Index.class);`

bidoubiwa: I prefer the second option for consistency, but you are maybe right. You can ignore this for the moment but I would like the opinion of @brunoocasali on that

bidoubiwa (approved):
LGTM 🔥🔥🔥🔥🔥

Nice work ✨✨✨✨

src/main/java/com/meilisearch/sdk/IndexesHandler.java
brunoocasali: My initial conception around `IndexesQuery`, `KeysQuery`, and so on, where to handle query parameters. If we are very strict, we can say that the `/indexes` path is related to the URL, not the **query parameters**, so it shouldn't be inside.

In any case, I don't think `/indexes` should be treated as handled inside of `IndexesQuery`, but if we use a `private constant` inside of `IndexesHandler` is enough.

If the path change in the future, we will have two places to change the tests, `IndexesQuery` and `IndexesHandler`.

PR #517, opened 2022-12-28
Keys changes

alallema:
# Pull Request

## What does this PR do?

### Changes

- `PATCH /keys/:uid_or_key` can update **only** `name` and `description` fields.
- [x] Add test for new `:name` attributes


src/main/java/com/meilisearch/sdk/KeysHandler.java
```diff
@@ -45,6 +50,23 @@ Results<Key> getKeys() throws MeilisearchException {
...
+    /**
+     * Retrieves keys from the client
+     *
+     * @param param accept by the keys route
+     * @return List of key instance
+     * @throws MeilisearchException if client request causes an error
+     */
+    Results<Key> getKeys(KeysQuery param) throws MeilisearchException {
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("keys")
```
brunoocasali, line 62: I left a similar comment in the documents PR, I think this should be done inside of the `KeysQuery`.

brunoocasali (changes requested):

src/main/java/com/meilisearch/sdk/KeysHandler.java
```diff
@@ -45,6 +46,18 @@ Results<Key> getKeys() throws MeilisearchException {
...
+    /**
+     * Retrieves keys from the client
+     *
+     * @param param accept by the keys route
+     * @return List of key instance
+     * @throws MeilisearchException if client request causes an error
+     */
+    Results<Key> getKeys(KeysQuery param) throws MeilisearchException {
+        Results<Key> result = httpClient.get(param.toQuery(param), Results.class, Key.class);
+        return result;
```
bidoubiwa, line 58: Does this work?

```suggestion
        return httpClient.get(param.toQuery(param), Results.class, Key.class);
```

src/main/java/com/meilisearch/sdk/KeysHandler.java
```diff
@@ -45,6 +46,18 @@ Results<Key> getKeys() throws MeilisearchException {
         return result;
     }
 
+    /**
+     * Retrieves keys from the client
+     *
+     * @param param accept by the keys route
+     * @return List of key instance
```
bidoubiwa, line 53: not a list anymore

src/main/java/com/meilisearch/sdk/KeysHandler.java
```diff
@@ -45,6 +46,18 @@ Results<Key> getKeys() throws MeilisearchException {
         return result;
     }
 
+    /**
+     * Retrieves keys from the client
```
bidoubiwa, line 50:
```suggestion
     * Get API keys
```

src/main/java/com/meilisearch/sdk/KeysHandler.java
```diff
@@ -63,11 +76,10 @@ Key createKey(Key options) throws MeilisearchException {
...
      * @param options String containing the options of the key
      * @return Key Instance
-     * @throws Exception if client request causes an error
+     * @throws MeilisearchException if client request causes an error
      */
-    Key updateKey(String key, Key options) throws Exception {
-        String urlPath = "/keys/" + key;
-        return httpClient.patch(urlPath, options, Key.class);
+    Key updateKey(String key, KeyUpdate options) throws MeilisearchException {
+        return httpClient.patch(new KeysQuery().toQuery(key), options, Key.class);
```
bidoubiwa, line 82: Same as for `getIndex` and `getIndexes`, should `updateKey` and `updatePrimaryKey` somehow work in similar ways?

src/main/java/com/meilisearch/sdk/KeysHandler.java
```diff
@@ -77,7 +89,6 @@ Key updateKey(String key, Key options) throws Exception {
      * @throws MeilisearchException if client request causes an error
      */
     void deleteKey(String key) throws MeilisearchException {
-        String urlPath = "/keys/" + key;
-        httpClient.delete(urlPath, String.class);
+        httpClient.delete(new KeysQuery().toQuery(key), String.class);
```
bidoubiwa, line 92: Same as for `getIndex` and `getIndexes`, should `deleteKey` and `deleteIndex` somehow work in similar ways?

src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
```diff
@@ -26,9 +29,20 @@ public String encode(Object o) throws MeilisearchException {
...
+            this.gson = builder.setDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'").create();
+
+            Key key = (Key) o;
+            if (key.getExpiresAt() == null) {
+                JsonElement jsonElement = gson.toJsonTree(o);
+                JsonObject jsonObject = jsonElement.getAsJsonObject();
+                jsonObject.add("expiresAt", JsonNull.INSTANCE);
+                o = jsonObject;
+                this.gson =
+                        builder.serializeNulls().setDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'").create();
```
bidoubiwa, line 44: what would be the value of `expiresAt` if we enter this condition?

bidoubiwa (changes requested):
Can you change the name of the PR with more context? For example

Update Keys API for Meilisearch v.28

----

Could you double check the documentation above the methods? I maybe forgot to suggest some changes

src/main/java/com/meilisearch/sdk/KeysHandler.java
alallema: No

src/main/java/com/meilisearch/sdk/KeysHandler.java
alallema: "Retrieves" is use in the definitions of all get method for documents, indexes and tasks so it should be changes in all of it. I can replace it by:
`Retrieves API keys from the client`
?

src/main/java/com/meilisearch/sdk/KeysHandler.java
```diff
@@ -29,8 +31,7 @@ public class KeysHandler {
      * @throws MeilisearchException if client request causes an error
      */
     Key getKey(String uid) throws MeilisearchException {
-        String urlPath = "/keys/" + uid;
-        return httpClient.get(urlPath, Key.class);
+        return httpClient.get(new KeysQuery().toQuery(uid), Key.class);
```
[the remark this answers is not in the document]
alallema: >I'm not sure if this is the best way to do it

How would you do it?
>should getIndex work the same way or should getKey work like getIndex?

Why should it work the same way?
I created a `toQuery()` method in `Key` class because this piece of code is repeated twice.

src/main/java/com/meilisearch/sdk/KeysHandler.java
alallema: Same answer as for `getKey()`

src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
alallema: the condition is:
`if (key.getExpiresAt() == null)`
So `expiredAt` should be null to enter in this condition and it should be encoded as null. That why the json builder is set like:
```java
            this.gson =
                        builder.serializeNulls().setDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'").create();
```
with `serializeNulls()`

src/main/java/com/meilisearch/sdk/KeysHandler.java
alallema: Actually it's working !!!

src/main/java/com/meilisearch/sdk/KeysHandler.java
alallema: I will open an issue about replacing all the retrieves by get

src/main/java/com/meilisearch/sdk/model/KeyUpdate.java
```diff
@@ -0,0 +1,23 @@
+package com.meilisearch.sdk.model;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+/**
+ * Data structure of Meilisearch response for a Key
```
bidoubiwa, line 8: Not sure this is the correct definition

src/main/java/com/meilisearch/sdk/model/KeysQuery.java
```diff
@@ -0,0 +1,35 @@
+package com.meilisearch.sdk.model;
+
+import com.meilisearch.sdk.http.URLBuilder;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+/**
+ * Data structure of a query parameter for keys route
```
bidoubiwa, line 9:
```suggestion
 * Data structure of the query parameters for the keys routes
```

src/test/java/com/meilisearch/integration/KeysTest.java
```diff
@@ -19,10 +21,6 @@ public class KeysTest extends AbstractIT {
     public void initialize() {
         this.setUp();
         this.setUpJacksonClient();
-    }
-
-    @AfterAll
```
bidoubiwa, line 24: You should keep an afterall since the beforeEach is not triggered after the last tests. Resulting in the indexes created in the last test not being deleted

src/test/java/com/meilisearch/integration/KeysTest.java
```diff
@@ -63,36 +62,104 @@ public void testClientGetKeysWithJacksonJsonHandler() throws Exception {
...
+        assertTrue(key instanceof Key);
+        assertNotNull(key.getKey());
+        assertNotNull(key.getActions());
+        assertNotNull(key.getIndexes());
+        assertNotNull(key.getDescription());
+        assertNotNull(key.getCreatedAt());
+        assertNotNull(key.getUpdatedAt());
+    }
+
+    /** Test Get Key */
```
bidoubiwa, line 128:
```suggestion
    /** Test Get Key when the key does not exist*/
```

bidoubiwa (changes requested):

bidoubiwa (approved):
LGTM 🔥🔥🔥🔥🔥 GREAT WORK

PR #518, opened 2022-12-28
Update documents API for Meilisearch v.28

alallema:
# Pull Request

## What does this PR do?

- `GET /documents/:uid`, `GET /documents` Add the possibility to reduce the body payload by using a query parameter called `fields` (previously called `attributesToRetrieve`), check the issue and the spec for the entire behavior.
- The documents objects now return wrap in a `Results`
- Create two separate methods for `GET /documents`:
    - `<T> Results<T> getDocuments(Class<T> targetClass) throws MeilisearchException` -> return an `Object`
    - `String getRawDocuments()` -> return a `String`
- Create two separate methods for `GET /documents/:uid`:
    - `<T> Results<T> getDocument(String identifier, Class<T> targetClass) throws MeilisearchException` -> return an `Object`
    - `String getRawDocument(String identifier)` -> return a `String`
- [x] Adapt DocumentsQuery naming following all the routes
- [x] Adapt and fix tests
- [x] Add test for `getRawDocument`
- [x] Add test for `getRawDocuments`

### DIFF
--- a/src/main/java/com/meilisearch/sdk/Documents.java
+++ b/src/main/java/com/meilisearch/sdk/Documents.java
@@ -39,7 +39,7 @@ class Documents {
      *
      * @param uid Partial index identifier for the requested documents
      * @param identifier ID of the document
-     * @param param accept by the documents route
+     * @param param accepted by the get document route
      * @param targetClass Class of the document returned
      * @return Object containing the requested document
      * @throws MeilisearchException if client request causes an error
--- a/src/main/java/com/meilisearch/sdk/Documents.java
+++ b/src/main/java/com/meilisearch/sdk/Documents.java
@@ -123,7 +123,7 @@ class Documents {
      * Retrieves the document as String at the specified index
      *
      * @param uid Partial index identifier for the requested documents
-     * @param param accept by the documents route
+     * @param param accepted by the documents route
      * @return Meilisearch API response
      * @throws MeilisearchException if an error occurs
      */
--- a/src/main/java/com/meilisearch/sdk/Documents.java
+++ b/src/main/java/com/meilisearch/sdk/Documents.java
@@ -96,7 +96,7 @@ class Documents {
      *
      * @param <T> Type of documents returned
      * @param uid Partial index identifier for the requested documents
-     * @param param accept by the documents route
+     * @param param accepted by the get documents route
      * @param targetClass Class of documents returned
      * @return Object containing the requested document
      * @throws MeilisearchException if the client request causes an error
--- a/src/main/java/com/meilisearch/sdk/Documents.java
+++ b/src/main/java/com/meilisearch/sdk/Documents.java
@@ -1,8 +1,9 @@
 package com.meilisearch.sdk;
 
-import static java.util.Collections.singletonList;
-
 import com.meilisearch.sdk.exceptions.MeilisearchException;
+import com.meilisearch.sdk.http.URLBuilder;
+import com.meilisearch.sdk.model.DocumentsQuery;
+import com.meilisearch.sdk.model.Results;
 import com.meilisearch.sdk.model.TaskInfo;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -21,82 +22,162 @@ protected Documents(Config config) {
...
+    <T> T getDocument(String uid, String identifier, DocumentsQuery param, Class<T> targetClass)
+            throws MeilisearchException {
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("indexes")
+                .addSubroute(uid)
+                .addSubroute("documents")
+                .addSubroute(identifier)
+                .addParameter("limit", param.getLimit())
+                .addParameter("offset", param.getOffset())
+                .addParameter("fields", param.getFields());
```
brunoocasali, line 62: This piece of code will repeat in all the occurrences where `DocumentsQuery` will be used. So why not create a method `toQuery` or something inside of the `DocumentsQuery`, then you can just call it.

`return httpClient.<T>get(param.toQuery(), targetClass);`

brunoocasali (changes requested):
I left a comment, but it applies to every occurrence of the queries :) 

I really think it will improve the complexity of the code :)

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -21,82 +22,162 @@ protected Documents(Config config) {
...
      *
+     * @param <T> Type of the document returned
      * @param uid Partial index identifier for the requested documents
      * @param identifier ID of the document
-     * @return String containing the requested document
+     * @param targetClass Class of the document returned
+     * @return Object containing the requested document
+     * @throws MeilisearchException if the client request causes an error
+     */
+    <T> T getDocument(String uid, String identifier, Class<T> targetClass)
```
brunoocasali, line 32: You have to also add support to `fields`, which allows the user to reduce the payload meilisearch will return like we have here: https://github.com/meilisearch/meilisearch-python/blob/a6dc0631103a495d109ff6b69e288193d1b22b73/tests/index/test_index_document_meilisearch.py#L71-L78.

brunoocasali: This can be used in the `getDocuments`, `getDocument`, `getRawDocuments`, and `getRawDocument`

brunoocasali (approved):

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -21,83 +22,117 @@ protected Documents(Config config) {
...
      * @throws MeilisearchException if the client request causes an error
      */
-    String getDocuments(String uid, int limit, int offset) throws MeilisearchException {
-        String urlQuery = "/indexes/" + uid + "/documents?limit=" + limit + "&offset=" + offset;
-        return httpClient.get(urlQuery, String.class);
+    <T> Results<T> getDocuments(String uid, Class<T> targetClass) throws MeilisearchException {
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("indexes").addSubroute(uid).addSubroute("documents");
+        String urlPath = urlb.getURL();
+        Results<T> documents = httpClient.<Results>get(urlPath, Results.class, targetClass);
```
bidoubiwa, line 91: Same as for Index and Keys, you can use the same toQuery method

bidoubiwa (changes requested):
Could you go over the different method documentations to update

- The explanations
- The `@param param`
- The return statement

I started suggesting on a lot of them

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -21,83 +22,117 @@ protected Documents(Config config) {
...
+     * Retrieves the document as String at the specified index
+     *
+     * @param uid Partial index identifier for the requested documents
+     * @return Meilisearch API response
+     * @throws MeilisearchException if an error occurs
+     */
+    String getRawDocuments(String uid) throws MeilisearchException {
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("indexes").addSubroute(uid).addSubroute("documents");
+        String urlPath = urlb.getURL();
```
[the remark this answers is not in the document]
alallema: Yes sorry I forgot to push my modification before your review ...

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -158,7 +197,12 @@ TaskInfo deleteDocument(String uid, String identifier) throws MeilisearchExcepti
      * @throws MeilisearchException if the client request causes an error
      */
     TaskInfo deleteDocuments(String uid, List<String> identifiers) throws MeilisearchException {
-        String urlPath = "/indexes/" + uid + "/documents/" + "delete-batch";
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("indexes")
+                .addSubroute(uid)
+                .addSubroute("documents")
+                .addSubroute("delete-batch");
+        String urlPath = urlb.getURL();
```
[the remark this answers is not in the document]
alallema: There already is a `toQuery` method as `public String toQuery(String uid)`
I can modify it to return the `urlb` instead of the URL directly to be able to add this:
`.addSubroute("delete-batch");`
But I don't think it will be better

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -129,10 +166,12 @@ TaskInfo addDocuments(String uid, String document, String primaryKey)
...
     TaskInfo updateDocuments(String uid, String document, String primaryKey)
             throws MeilisearchException {
-        String urlPath = "/indexes/" + uid + "/documents";
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("indexes").addSubroute(uid).addSubroute("documents");
         if (primaryKey != null) {
-            urlPath += "?primaryKey=" + primaryKey;
+            urlb.addParameter("primaryKey", primaryKey);
         }
+        String urlPath = urlb.getURL();
```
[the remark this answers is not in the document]
bidoubiwa: I think it would be: 
```
    public String toQuery(String uid, String identifier, String primaryKey) {
```
Which I don't think exists yet in the toQuery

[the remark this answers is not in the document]
alallema: > I think it would be:
`    public String toQuery(String uid, String identifier, String primaryKey) {`

It could but we don't use `identifier`

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -111,10 +146,12 @@ String getDocuments(String uid, int limit, int offset, List<String> attributesTo
      */
     TaskInfo addDocuments(String uid, String document, String primaryKey)
             throws MeilisearchException {
-        String urlQuery = "/indexes/" + uid + "/documents";
+        URLBuilder urlb = new URLBuilder();
+        urlb.addSubroute("indexes").addSubroute(uid).addSubroute("documents");
```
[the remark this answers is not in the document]
alallema: This is the same case as in the line 169-174

src/main/java/com/meilisearch/sdk/Documents.java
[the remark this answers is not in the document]
alallema: > @brunoocasali what do you think?
Same

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -19,85 +20,111 @@ protected Documents(Config config) {
...
-            attributesToRetrieve = singletonList("*");
-        }
+        return httpClient.<Results>get(param.toQuery(uid, param), Results.class, targetClass);
+    }
+
+    /**
+     * Retrieves the document as a string from the specified index
+     *
+     * @param uid Partial index identifier for the requested documents
+     * @return Meilisearch API response
```
bidoubiwa, line 111:
```suggestion
     * @return String containing the requested documents
```

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -19,85 +20,111 @@ protected Documents(Config config) {
...
-                        + "&attributesToRetrieve="
-                        + attributesToRetrieveCommaSeparated;
-
-        return httpClient.get(urlQuery, String.class);
+    /**
+     * Retrieves the document as String at the specified index
+     *
+     * @param uid Partial index identifier for the requested documents
+     * @param param accepted by the documents route
+     * @return Meilisearch API response
```
bidoubiwa, line 123:
```suggestion
     * @return String containing the requested document
```

src/main/java/com/meilisearch/sdk/Documents.java
```diff
@@ -19,85 +20,111 @@ protected Documents(Config config) {
...
-                        + "/documents?limit="
-                        + limit
-                        + "&offset="
-                        + offset
-                        + "&attributesToRetrieve="
-                        + attributesToRetrieveCommaSeparated;
-
-        return httpClient.get(urlQuery, String.class);
+    /**
+     * Retrieves the document as String at the specified index
```
bidoubiwa, line 119:
```suggestion
     * Retrieves the document as String from the specified index
```

bidoubiwa (approved):

PR #524, opened 2023-01-10
Refactor handlers for the usage of URLBuilder

alallema:
- Creation of the class `DocumentQuery` to getting one document.

**Notes:**
- A function has been created in each handler to avoid instantiating a single `new URLBuilder()` in the class parameters.

### DIFF
--- a/src/main/java/com/meilisearch/sdk/http/URLBuilder.java
+++ b/src/main/java/com/meilisearch/sdk/http/URLBuilder.java
@@ -52,8 +52,9 @@ public class URLBuilder {
         return this;
     }
 
-    public String addQuery(String query) {
-        return routes.toString() + params.toString() + query;
+    public URLBuilder addQuery(String query) {
+        this.params.append(query);
+        return this;
     }
 
     private URLBuilder addSeparator() {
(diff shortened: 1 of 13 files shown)


src/main/java/com/meilisearch/sdk/http/URLBuilder.java
```diff
@@ -52,6 +52,10 @@ public URLBuilder addParameter(String parameter, String[] value) {
         return this;
     }
 
+    public String addQuery(String query) {
```
brunoocasali, line 55: When I suggested in the meeting to create this method, I was expecting this:

```java
public URLBuilder addQuery(String query) {
  this.addSeparator();
  this.params.append(query);

  return this;
}
```

brunoocasali (approved):
I left a suggestion but feel free to ignore it.

Nice work!

src/main/java/com/meilisearch/sdk/http/URLBuilder.java
alallema: Yes, I was wondering if it was better to send the object back, and then we could do the following:
`urlb.addQuery(param.ToQuery).getURL();`

brunoocasali: I think it is better to change to the `URLBuilder` because `addQuery` gives me the feeling that I could keep appending stuff, like: `addSubroute(...).addSeparator()...`.

alallema: Indeed

Issue #455, opened 2022-10-13
Make the SearchRequest class a builder


alallema:
**Description**
The class `SearchRequest` has many attributes and it's difficult to create initializers that will suit every possible case. It will be better to use [builder](https://projectlombok.org/features/Builder) instead.

**Basic example**
```java
SearchRequest.builder()
    .q("")
    .offset(10)
    .limit(200)
    .cropLength(5)
    .cropMarker("…")
    .build();
```

**Notes**
This issue should be implemented before the resolution of #420

PR #543, opened 2023-02-15
Add a builder for SearchRequest class

alallema:
The `SearchRequest` class has been modified to simplify its use, see #455

- Removing all the constructors with multiple arguments on the Class
- Add the possibility to use a `builder()`
```
SearchRequest.builder()
    .q("")
    .offset(10)
    .limit(200)
    .cropLength(5)
    .cropMarker("…")
    .build();
```
- Keep the possibility to initialize the Class like:
```
new SearchRequest("")
    .setOffset(10)
    .setLimit(200)
    .setCropLength(5)
    .setCropMarker("…");
```

### DIFF
--- a/src/main/java/com/meilisearch/sdk/SearchRequest.java
+++ b/src/main/java/com/meilisearch/sdk/SearchRequest.java
@@ -70,10 +70,10 @@ public class SearchRequest {
         JSONObject jsonObject =
                 new JSONObject()
                         .put("q", this.q)
-                        .put("offset", this.offset == null ? null : this.offset)
-                        .put("limit", this.limit == null ? null : this.limit)
+                        .put("offset", this.offset)
+                        .put("limit", this.limit)
                         .put("attributesToRetrieve", this.attributesToRetrieve)
-                        .put("cropLength", this.cropLength == null ? null : this.cropLength)
+                        .put("cropLength", this.cropLength)
                         .put("cropMarker", this.cropMarker)
                         .put("highlightPreTag", this.highlightPreTag)
                         .put("highlightPostTag", this.highlightPostTag)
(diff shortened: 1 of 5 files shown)


src/main/java/com/meilisearch/sdk/SearchRequest.java
```diff
@@ -356,10 +70,10 @@ public String toString() {
         JSONObject jsonObject =
                 new JSONObject()
                         .put("q", this.q)
-                        .put("offset", this.offset)
-                        .put("limit", this.limit)
+                        .put("offset", this.offset == null ? null : this.offset)
```
brunoocasali, line 73: Why do we need this check? If the `this.offset` is `null`, can we just keep `.put("offset", this.offset)` ?

brunoocasali: The same for the other occurrences!

src/main/java/com/meilisearch/sdk/SearchRequest.java
```diff
@@ -1,36 +1,40 @@
...
     private String[] attributesToHighlight;
     private String[] filter;
     private String[][] filterArray;
-    private boolean showMatchesPosition;
+    private Boolean showMatchesPosition;
     private String[] facets;
     private String[] sort;
 
-    /** Empty SearchRequest constructor */
-    public SearchRequest() {}
```
brunoocasali, line 32: I think you can make it return `this`, and then you'll enable a new version (my preferred):

```java
new SearchRequest()
  .setQuery("my query") // or I think it will be just `setQ` - good as well.
  .setOffset(200)
  .setLimit(900);
```

WDYT?

brunoocasali (approved):
I'll let you my approval. This is a fantastic improvement! Congrats @alallema you rocks!! 🎉

src/main/java/com/meilisearch/sdk/SearchRequest.java
alallema: I modifying a test with this syntax `toStringQueryAndOffset`

brunoocasali (approved):
LGTM!

alallema:
bors merge

PR #547, opened 2023-02-16
Add cancel task api for v0.30.0

alallema:
Add the cancel task api see spec:  https://github.com/meilisearch/specifications/pull/195

- [x] Implement `CancelTasks()` in the client
  - params: `CancelTasksQuery`
  - returns: `TaskInfo`
- [x] Tests cancelation on query parameters

### DIFF
--- a/src/main/java/com/meilisearch/sdk/model/CancelTasksQuery.java
+++ b/src/main/java/com/meilisearch/sdk/model/CancelTasksQuery.java
@@ -1,0 +1,41 @@
+package com.meilisearch.sdk.model;
+
+import com.meilisearch.sdk.http.URLBuilder;
+import java.util.Date;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+/**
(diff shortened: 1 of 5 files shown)


src/main/java/com/meilisearch/sdk/model/CancelTasksQuery.java
```diff
@@ -0,0 +1,41 @@
...
+    private Date afterEnqueuedAt;
+    private Date beforeStartedAt;
+    private Date afterStartedAt;
+
+    public CancelTasksQuery() {}
+
+    public String toQuery() {
+        URLBuilder urlb =
+                new URLBuilder()
+                        .addParameter("statuses", this.getStatuses())
```
brunoocasali, line 32:
```suggestion
                        .addParameter("uids", this.getUids())
                        .addParameter("statuses", this.getStatuses())
```

brunoocasali (changes requested):
The `uids` is missing in the final URL! The rest is perfect 🍐

PR #548, opened 2023-02-16
Add delete task api for v0.30.0

alallema:
Add the delete task api see spec:  https://github.com/meilisearch/specifications/pull/198

- [x] Implement `DeleteTasks()` in the client
  - params: `DeleteTasksQuery`
  - returns: `TaskInfo`
- [x] Tests deletion on query parameters

### DIFF
--- a/src/main/java/com/meilisearch/sdk/model/DeleteTasksQuery.java
+++ b/src/main/java/com/meilisearch/sdk/model/DeleteTasksQuery.java
@@ -1,0 +1,47 @@
+package com.meilisearch.sdk.model;
+
+import com.meilisearch.sdk.http.URLBuilder;
+import java.util.Date;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+/**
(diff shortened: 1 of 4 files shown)


src/main/java/com/meilisearch/sdk/model/DeleteTasksQuery.java
```diff
@@ -0,0 +1,47 @@
...
+    private Date afterStartedAt;
+    private Date beforeFinishedAt;
+    private Date afterFinishedAt;
+
+    public DeleteTasksQuery() {}
+
+    public String toQuery() {
+        URLBuilder urlb =
+                new URLBuilder()
+                        .addParameter("statuses", this.getStatuses())
```
brunoocasali, line 35:
```suggestion
                        .addParameter("uids", this.getUids())
                        .addParameter("statuses", this.getStatuses())
```

brunoocasali (changes requested):
Same as the `CancelTasksQuery` PR :) Everything else is good!

PR #554, opened 2023-02-21
Add swap indexes api for v0.30.0

alallema:
As [per the specification](https://github.com/meilisearch/specifications/pull/192)

## SWAP INDEXES

- [x] create `swapIndexesParams` type
- [x] create `swapIndexes` method in the client
    - params: TasksQuery
    - returns: TaskInfo
- [x] test to swap on indexes:
- [x] add `swaps` in task details

### DIFF
--- a/src/test/java/com/meilisearch/integration/ClientTest.java
+++ b/src/test/java/com/meilisearch/integration/ClientTest.java
@@ -11,6 +11,7 @@ import com.meilisearch.sdk.Index;
 import com.meilisearch.sdk.exceptions.MeilisearchApiException;
 import com.meilisearch.sdk.model.IndexesQuery;
 import com.meilisearch.sdk.model.Results;
+import com.meilisearch.sdk.model.SwapIndexesParams;
 import com.meilisearch.sdk.model.Task;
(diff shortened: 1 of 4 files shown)


src/test/java/com/meilisearch/integration/ClientTest.java
```diff
@@ -201,6 +202,37 @@ public void testDeleteIndex() throws Exception {
         assertThrows(MeilisearchApiException.class, () -> client.getIndex(indexUid));
     }
 
+    /** Test swapIndexes */
+    @Test
+    public void testSwapIndexes() throws Exception {
```
brunoocasali, line 207: I appreciate the effort to test it, but it could be more straightforward. Just the assertion to make sure you could request using the correct indexes and retrieve the type of the task `indexSwap` would have been ok.

brunoocasali (approved):
LGTM! 🥇

src/test/java/com/meilisearch/integration/ClientTest.java
alallema: I knew you were going to say that when I wrote the test but I couldn't help myself 🤭

PR #561, opened 2023-02-27
Fix matching strategy enum usage and add test

alallema:
This PR corrects the use of the `matchingStrategy` enum which sends `all` and `last` in upper case instead of lower case.


brunoocasali (approved):

alallema:
bors merge

PR #566, opened 2023-03-02
feat: created Task Status enum.

### LINKED ISSUE #556: Create an enum to handle the Task Status
**Description**
Currently, the Task status in `TaskInfo` is a `String`
```java
public class TaskInfo {
    protected String status = "";
    protected int taskUid = 0;
    protected String indexUid = "";
    protected String type = null;
    protected Date enqueuedAt = null;

    public TaskInfo() {}
}
```
But the status could only be one of this value: `enqueued`, `processing`, `succeeded`, `failed`, `canceled`.
It could be great to create an [enum](https://www.baeldung.com/a-guide-to-java-enums) to handle this like in the [.NET repository](https://github.com/meilisearch/meilisearch-dotnet/blob/0acea9667cbdec92e04c197422b4c8d1f88a15e5/src/Meilisearch/TaskInfo.cs#L96).

TheClerici:
# Pull Request
Task status enum on model package, added on task and taskInfo and modified some tests.

## Related issue
Fixes #556 

## What does this PR do?
* TaskStatus enum added in model directory
* String status in the Task and TaskInfo class exchanged with new enum
* Some tests rewritten with the new use of the enum

## PR checklist

I have a question. While changing some test I found out that when trying to do .waitForTask a NullPointerException was happening, even tho this not affect directly to what I'm implementing I don't wanna move something that can make it break, I also built  everything and  tested with ./gradlew test and lint. This is a draft so maybe you can check it out and let me know what you think

### DIFF
--- a/src/main/java/com/meilisearch/sdk/TasksHandler.java
+++ b/src/main/java/com/meilisearch/sdk/TasksHandler.java
@@ -13,8 +13,6 @@ import java.util.Date;
  */
 public class TasksHandler {
     private final HttpClient httpClient;
-    public static final String SUCCEEDED = "succeeded";
-    public static final String FAILED = "failed";
 
     /**
      * Creates and sets up an instance of Task to simplify MeiliSearch API calls to manage tasks
@@ -140,11 +138,12 @@ public class TasksHandler {
      */
     void waitForTask(int taskUid, int timeoutInMs, int intervalInMs) throws MeilisearchException {
         Task task;
-        TaskStatus status = TaskStatus.ENQUEUED;
+        TaskStatus status = null;
         long startTime = new Date().getTime();
         long elapsedTime = 0;
 
-        while (!status.equals(SUCCEEDED) && !status.equals(FAILED)) {
+        while (status == null
+                || (!status.equals(TaskStatus.SUCCEEDED) && !status.equals(TaskStatus.FAILED))) {
             if (elapsedTime >= timeoutInMs) {
                 throw new MeilisearchTimeoutException();
             }
--- a/src/main/java/com/meilisearch/sdk/TasksHandler.java
+++ b/src/main/java/com/meilisearch/sdk/TasksHandler.java
@@ -3,12 +3,7 @@ package com.meilisearch.sdk;
 import com.meilisearch.sdk.exceptions.MeilisearchException;
 import com.meilisearch.sdk.exceptions.MeilisearchTimeoutException;
 import com.meilisearch.sdk.http.URLBuilder;
-import com.meilisearch.sdk.model.CancelTasksQuery;
-import com.meilisearch.sdk.model.DeleteTasksQuery;
-import com.meilisearch.sdk.model.Task;
-import com.meilisearch.sdk.model.TaskInfo;
-import com.meilisearch.sdk.model.TasksQuery;
-import com.meilisearch.sdk.model.TasksResults;
+import com.meilisearch.sdk.model.*;
 import java.util.Date;
 
 /**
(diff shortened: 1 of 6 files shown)


src/test/java/com/meilisearch/integration/TasksTest.java
```diff
@@ -224,7 +219,7 @@ public void testClientDeleteTaskWithMultipleFilters() throws Exception {
 
         assertTrue(task instanceof TaskInfo);
         assertNotNull(task.getStatus());
-        assertNotEquals("", task.getStatus());
+        assertNotEquals(null, task.getStatus());
```
alallema, line 222:
```suggestion
```
No need for this second line it tests exactly the same thing as the line above

src/main/java/com/meilisearch/sdk/TasksHandler.java
```diff
@@ -3,12 +3,7 @@
 import com.meilisearch.sdk.exceptions.MeilisearchException;
 import com.meilisearch.sdk.exceptions.MeilisearchTimeoutException;
 import com.meilisearch.sdk.http.URLBuilder;
-import com.meilisearch.sdk.model.CancelTasksQuery;
-import com.meilisearch.sdk.model.DeleteTasksQuery;
-import com.meilisearch.sdk.model.Task;
-import com.meilisearch.sdk.model.TaskInfo;
-import com.meilisearch.sdk.model.TasksQuery;
-import com.meilisearch.sdk.model.TasksResults;
```
alallema, line 11: 👍

alallema (commented):
Thanks, @TheClerici, for this PR ❤️!
> I found out that when trying to do .waitForTask a NullPointerException was happening

Yes, The enum implementation is great, as well as the modification of the tests. But the modification of the `waitForTask` has to be fixed, I add some suggestions.

TheClerici:
Perfect, I will continue today then. I will modify the line you told me on *testCreateDump()*. Sorry but I don't see clearly what to do how to modify *waitForTask* to make it work on the tests, maybe you can give me a hint. I will follow the suggestions you made too and change them.

alallema:
> Perfect, I will continue today then. I will modify the line you told me on _testCreateDump()_. Sorry but I don't see clearly what to do how to modify _waitForTask_ to make it work on the tests, maybe you can give me a hint. I will follow the suggestions you made too and change them.

Hi @TheClerici,
Sorry if I wasn't clear enough, you were right to modify the `waitForTask` method but the reason why it fails is that now the status returned is a `TaskStatus` type and no longer a `String` so the condition must be modified as
```java
        while (status == null
                || (!status.equals(TaskStatus.SUCCEEDED) && !status.equals(TaskStatus.FAILED))) {
```
The condition must also include a check that the status is not null because it is better to initialize it to null rather than to `TaskStatus.ENQUEUED` so the initialization becomes:
```java
        TaskStatus status = null;
```
And theses both lines at the beginning of the file could be removed:
```java
    public static final String SUCCEEDED = "succeeded";
    public static final String FAILED = "failed";
```
Finally the `waitForTask` method become:
```java
    void waitForTask(int taskUid, int timeoutInMs, int intervalInMs) throws MeilisearchException {
        Task task;
        TaskStatus status = null;
        long startTime = new Date().getTime();
        long elapsedTime = 0;

        while (status == null
                || (!status.equals(TaskStatus.SUCCEEDED) && !status.equals(TaskStatus.FAILED))) {
        ...
        }
    }
```

TheClerici:
perfect! Thanks! I will work on it

TheClerici:
Hello! @alallema I followed the suggestion you made and I believe it is ready, lmk! Thanks you so much for all the feedback.

alallema (approved):
Thanks, @TheClerici for this PR 🎉 
LGTM!
Just FYI, we encourage everyone to make a PR from a new branch created in their fork-repository instead of doing it from their own main, it is a good practice! For example, if your PR was not merged you would have had a broken main branch compared to ours.
See here our contribution guidelines.
I'm saying this for the next time, let's keep it like this for this PR. :wink:

alallema:
bors merge

TheClerici:
Thanks!! I will have that in mind for next time! I forgot about it, I made it like that the first time hahaha. I appreciate all the help! Have a nice day :D

PR #574, opened 2023-03-17
Updated Readme.md - fixed maven-dependency

KaaNee:
Changed dependecy-type from "pom" to "compile" in readme - when using pom it will only add the dependecies declared in the meilisearch-java-pom, but not the SDK itself. Below the definition for gradle uses "implementation", this means "compile" in Maven-Slang. To afford the same goal (use the SDK in the project), it should be compile, not only pom.

# Pull Request

## Related issue
none

## What does this PR do?
- changes documentation for using SDK with maven


KaaNee:
my mistake, ignore it.

PR #643, opened 2023-08-21
Fix possible unintended definition of BasicResponse.create

### LINKED ISSUE #642: Possible unintended definition of BasicResponse.create
I noticed when re-writing the tests in [this PR](https://github.com/meilisearch/meilisearch-java/pull/638) that the type signature of `BasicResponse.create` appears to conflict with the way in which it is used in one of the tests. This lead me to believe that the intention behind `BasicResponse.create` is a little different to what it is currently doing. This is best illustrated by showing some code:

```java
    public <T> HttpResponse<T> create(
            HttpResponse<T> httpResponse, Class<T> targetClass, Class<?>... parameters) {
        try {
            T content = this.jsonHandler.decode(httpResponse.getContent(), targetClass, parameters);

            return new HttpResponse<T>(
                    httpResponse.getHeaders(),
                    httpResponse.getStatusCode(),
                    httpResponse.getContent() == null ? null : content);
        } catch (Exception e) {
            throw new RuntimeException(e);
        }
    }
```

This is the code that I'm talking about. The generic parameter `T` here is the type that the _returned_ `HttpResponse` is generic over. The first argument is a `HttpResponse<T>` which suggests that whatever `T` is passed in as the type parameter to first argument, should be the exact same `T` that is returned.

If this was the intention then there would be no need for the second argument of type `Class<T>` since `T` could be inferred from whatever was passed in as the first argument.

However, the second parameter `Class<T> targetClass` is present, and presumably is there so that the caller can explicitly say what `T` should be, suggeseting that there's perhaps another type expected here (probably the return type).

This is reinforced by the associated test for this method:

```java
        String content =
                "{ \"uid\": 0, \"indexUid\": \"\", \"status\": \"\", \"type\": null, \"details\": null, \"duration\": \"\", \"enqueuedAt\": null, \"startedAt\": null, \"finishedAt\": null}";
        HttpResponse response = new HttpResponse(null, 0, content);
        HttpResponse<Task> httpResponse = basicResponse.create(response, Task.class);
```

As you can see, there's no type argument when defining `HttpResponse response = ...`. It is clear that the type that is actually constructed is `HttpResponse<String>` because the third argument of the constructor (`content`) is of type `String`. However, if we change the above test to this:

```java
        String content =
                "{ \"uid\": 0, \"indexUid\": \"\", \"status\": \"\", \"type\": null, \"details\": null, \"duration\": \"\", \"enqueuedAt\": null, \"startedAt\": null, \"finishedAt\": null}";
        HttpResponse<String> response = new HttpResponse(null, 0, content); // <--- added <String> here
        HttpResponse<Task> httpResponse = basicResponse.create(response, Task.class);
```

We get a compiler error, because the signature of `BasicResponse.create` is expecting that the returned `HttpResponse<T>` and the first argument `HttpResponse<T>` to have the same `T`, but since `String` is not the same as `Task`, it complains.

If my assumptions are correct here, and the intention of this method is to be able to take a `HttpResponse` of some type and transform it into a `HttpResponse` of some other type, then the fix for this is pretty simple. I can submit a PR to do this and you are free to accept it if it does the trick for you 😄

jrhenderson1988:
# Pull Request

## Related issue
Fixes #642 

## What does this PR do?
- Fixes a minor issue with `BasicResponse.create`, allowing the second argument to declare the type parameter of the returned `HttpResponse`

## PR checklist

### DIFF
--- a/src/main/java/com/meilisearch/sdk/http/response/BasicResponse.java
+++ b/src/main/java/com/meilisearch/sdk/http/response/BasicResponse.java
@@ -9,12 +9,12 @@ public class BasicResponse {
         this.jsonHandler = jsonHandler;
     }
 
-    public <T> HttpResponse<T> create(
-            HttpResponse<T> httpResponse, Class<T> targetClass, Class<?>... parameters) {
+    public <T, R> HttpResponse<R> create(
+            HttpResponse<T> httpResponse, Class<R> targetClass, Class<?>... parameters) {
         try {
-            T content = this.jsonHandler.decode(httpResponse.getContent(), targetClass, parameters);
+            R content = this.jsonHandler.decode(httpResponse.getContent(), targetClass, parameters);
 
-            return new HttpResponse<T>(
+            return new HttpResponse<>(
                     httpResponse.getHeaders(),
                     httpResponse.getStatusCode(),
                     httpResponse.getContent() == null ? null : content);
--- a/src/test/java/com/meilisearch/sdk/http/BasicResponseTest.java
+++ b/src/test/java/com/meilisearch/sdk/http/BasicResponseTest.java
@@ -68,11 +68,10 @@ public class BasicResponseTest {
     }
 
     @Test
-    @SuppressWarnings("unchecked")
     void contentClass() {
         String content =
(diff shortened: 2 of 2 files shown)


brunoocasali:
Can you take a look at his findings @alallema? (the issue as well)

alallema (approved):
Thanks again @jrhenderson1988 for this PR
LGTM! 🎉

alallema:
bors merge

PR #667, opened 2023-10-01
- use varargs instead of string arrays

singlyfy:
# Pull Request

## What does this PR do?

it replaces String[] arguments with varargs (String ... foo). It simplifies method calls and it should be backwards compatible. 
NOTE:  
This pull requests only changes method signature, additional validation should be added in places where at least one value is needed.

### DIFF
--- a/src/main/java/com/meilisearch/sdk/Config.java
+++ b/src/main/java/com/meilisearch/sdk/Config.java
@@ -57,7 +57,7 @@ public class Config {
      * @param apiKey API key to pass to the header of requests sent to Meilisearch
      * @param clientAgents List of customized agents to be passed to User-Agent header.
      */
-    public Config(String hostUrl, String apiKey, String[] clientAgents) {
+    public Config(String hostUrl, String apiKey, String... clientAgents) {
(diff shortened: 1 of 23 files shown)


curquiza:
@singlyfy thanks for the PR
Can you fix the tests (integration/unit tests)? cf the CI

Also, is this PR breaking for the users? I really ask, I don't necessarily refuse breaking PRs since the library is not stable yet, if it improves the user experience.
I'm not a Java developer, that's why I ask.

brunoocasali:
> String[] arguments with varargs (String ... foo)

The change should not be breaking @curquiza I tried this sample:

```java
import java.util.Arrays;

public class MyClass {
    public static void main(String args[]) {
      opa("opa", "123");
      opa(new String[] {"opa","data"});
    }
    
    public static void opa(String... arg) {
        System.out.println(Arrays.toString(arg));
    }
}
```

and compiled + execute nicely!

curquiza:
Thank you @brunoocasali 

Still needs @singlyfy to fix the tests and we are good

curquiza:
Closing for lack of activity

PR #690, opened 2023-11-25
Improve GSON JSON handler, moving Key specific logic

### LINKED ISSUE #487: Improve Json Handler by removing the specific serialization of `Key.class`
**Description**:

Currently when serializing data via the `encode` method of the JSON handler the class `Key` class is treated separately because of the date format.
We should find a solution so that all the dates are well managed in the JSON handler

**Condition to remove**:
```java
    if (o != null && o.getClass() == Key.class)
```

jrhenderson1988:
# Pull Request

## Related issue
Fixes #487

## What does this PR do?
- Removes Key specific logic from `GsonJsonHandler` in favour of creating a type adapter for serializing and deserializing `Key`
- Removes a constructor in `GsonJsonHandler` which seemingly is only used for testing (a _spy_ was created in one of the unit tests, recording actions taken - this unit test has been updated). This specific constructor allowed a `Gson` instance to be passed in, which would allow bypassing the addition of the type adapter, which would break the `Key` specific functionality just added.
  Depending on how you view this, it could be seen as a breaking change, if you consider the `GsonJsonHandler` to be part of the public API. However, I would argue that it should not be and I don't see another way (though I may well be wrong) to feasibly move the `Key` logic out of this class.
- Moved the `@Setter` annotation on `Key` to class level, which results in a setter being generated for every field. Previously, setters were only generated for some fields. If this is a problem, then I think the way to get around this is to use reflection to set fields in the type adapter. This is less ideal but it's possible to do. I believe the Gson library itself (Jackson too) uses reflection to set fields anyway. I am happy to make this change and roll back the changes made to `Key` if this is preferable.
- Added many more unit tests to `GsonJsonHandlerTest` related to encoding/decoding `Key`, and updates some earlier tests. These tests were added to compare the existing functionality (before my changes - with the `Key` specific logic still in place) with the changes that I have made, to ensure that they are compatible and that there are no differences/breakages. These tests could theoretically be committed first and they should pass against the `GsonJsonHandler` as it is now (before this PR).

## PR checklist

### DIFF
--- a/src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
+++ b/src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
@@ -13,14 +13,13 @@ import com.meilisearch.sdk.exceptions.MeilisearchException;
 import com.meilisearch.sdk.model.Key;
 
 public class GsonJsonHandler implements JsonHandler {
-    private Gson gson;
+
+    private final Gson gson;
 
     public GsonJsonHandler() {
-        this.gson = new Gson();
-    }
-
-    public GsonJsonHandler(Gson gson) {
-        this.gson = gson;
+        GsonBuilder builder = new GsonBuilder();
+        builder.registerTypeAdapter(Key.class, new GsonKeyTypeAdapter());
+        this.gson = builder.create();
     }
 
     @Override
@@ -28,22 +27,7 @@ public class GsonJsonHandler implements JsonHandler {
         if (o != null && o.getClass() == String.class) {
             return (String) o;
         }
-        // TODO: review later
-        // This is a workaround to encode the Key class properly
-        if (o != null && o.getClass() == Key.class) {
-            GsonBuilder builder = new GsonBuilder();
-            this.gson = builder.setDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'").create();
 
-            Key key = (Key) o;
-            if (key.getExpiresAt() == null) {
-                JsonElement jsonElement = gson.toJsonTree(o);
-                JsonObject jsonObject = jsonElement.getAsJsonObject();
-                jsonObject.add("expiresAt", JsonNull.INSTANCE);
-                o = jsonObject;
-                this.gson =
-                        builder.serializeNulls().setDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'").create();
-            }
-        }
         try {
             return gson.toJson(o);
         } catch (Exception e) {
@@ -54,7 +38,7 @@ public class GsonJsonHandler implements JsonHandler {
     @Override
     @SuppressWarnings("unchecked")
     public <T> T decode(Object o, Class<?> targetClass, Class<?>... parameters)
-            throws MeilisearchException {
+        throws MeilisearchException {
         if (o == null) {
             throw new JsonDecodingException("Response to deserialize is null");
         }
--- a/src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
+++ b/src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
@@ -35,7 +35,7 @@ public class GsonJsonHandler implements JsonHandler {
     @Override
     @SuppressWarnings("unchecked")
     public <T> T decode(Object o, Class<?> targetClass, Class<?>... parameters)
-        throws MeilisearchException {
+            throws MeilisearchException {
         if (o == null) {
             throw new JsonDecodingException("Response to deserialize is null");
         }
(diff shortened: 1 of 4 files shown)


src/main/java/com/meilisearch/sdk/json/GsonJsonHandler.java
```diff
@@ -13,37 +10,21 @@
...
-    public GsonJsonHandler() {
-        this.gson = new Gson();
-    }
+    private final Gson gson;
 
-    public GsonJsonHandler(Gson gson) {
-        this.gson = gson;
+    public GsonJsonHandler() {
+        GsonBuilder builder = new GsonBuilder();
+        builder.registerTypeAdapter(Key.class, new GsonKeyTypeAdapter());
```
brunoocasali, line 18: Can we do that "automatically" only when are about to serialize a Key?

jrhenderson1988: We could, but that would mean doing a check in the `encode(Object o)` to ensure the type of `o` is `Key`. I figured that one of the primary objectives was to remove the key specific checks and logic.

The primary purpose of a "type adapter", is that we can tell Gson, that when it comes to encoding or decoding a specific type (in this case, `Key.class` as the first argument), we can control what happens during that process. We're not really affecting how the rest of the Gson library works for any type, other than `Key`.

brunoocasali: Oh, I got it! That's cool :D

src/main/java/com/meilisearch/sdk/json/GsonKeyTypeAdapter.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @Override
+    public Key read(JsonReader reader) throws IOException {
+        return nullOrElse(
+                reader,
+                r -> {
+                    Key key = new Key();
+                    readStartObject(r);
+                    while (r.peek() != JsonToken.END_OBJECT) {
```
brunoocasali, line 62: Are you really obligated to define all the keys of the object? Can't you just change the ones you don't like the default shape?

brunoocasali: Can we have something more dynamic, because in the future releases that we change/include new items into the keys object we will have to remember to update this file as well :)

src/main/java/com/meilisearch/sdk/model/Key.java
```diff
@@ -13,29 +13,25 @@
  * @see <a href="https://www.meilisearch.com/docs/reference/api/keys">API specification</a>
  */
 @Getter
+@Setter
```
brunoocasali, line 16: I understand the move, but now the user can redefine the uid/key, which they are not allowed to (directly). Do you think is there any other way to handle that?

brunoocasali (commented):
Just a set of comments to understand the better the change...

But I'm okay to merge it anyway!

CC: @curquiza

src/main/java/com/meilisearch/sdk/model/Key.java
jrhenderson1988: Yeah, the way to do this would be to use runtime reflection, I think. I wrote a bit about this in my third bullet in my PR description. If you want me to do this, let me know.

src/main/java/com/meilisearch/sdk/json/GsonKeyTypeAdapter.java
jrhenderson1988: > Are you really obligated to define all the keys of the object? Can't you just change the ones you don't like the default shape?

I couldn't find a way to do that, unfortunately. That is what I wanted to do at first. But it seems that when you define a type adapter like this one, Gson will just call your code and I couldn't find a way to fall back to some kind of default implementation for some parts.

I also tried just writing a serializer but I couldn't get it working (I can't remember why now as it's been a little while since I worked on this)

> Can we have something more dynamic, because in the future releases that we change/include new items into the keys object we will have to remember to update this file as well :)

I totally understand where you're coming from. I don't think this approach is ideal. I actually prefer the approach that you already have in the project (though there's a few small things I'd change) because I think it allows for a lighter touch in maintainence. However, it seems that the purpose behind this ticket is to remove that Key specific logic from the encode method in the `GsonJsonHandler` and this felt like the most logical way to do that.

I can always spend some more time trying out some different approaches, if you'd prefer a different method that is more maintanable?

brunoocasali: No no, I think it is ok the way it is, I just asked to ensure I extract every bit of information ;)

src/main/java/com/meilisearch/sdk/model/Key.java
brunoocasali: It will not be a huge deal in any case. No need to change!

brunoocasali (approved):
LGTM!

curquiza:
thank you @jrhenderson1988 🙏 

bors merge

PR #734, opened 2024-05-20
fix WaitForTask timeout in case of a task cancellation

### LINKED ISSUE #725: WaitForTask timeout in case of a task cancellation
**Description**
Hello, I was looking for bugs in the SDK and realized that when we call `WaitForTask`:
https://github.com/meilisearch/meilisearch-java/blob/969d576d131c48c54219503ebd4c689a3207b737/src/main/java/com/meilisearch/sdk/TasksHandler.java#L145-L151

This condition is wrong: 
```java
status == null
                || (!status.equals(TaskStatus.SUCCEEDED) && !status.equals(TaskStatus.FAILED))
```

If the task gets canceled, then the loop hangs forever until it times out.
Instead of checking if the task has succeeded or failed, we should instead use this condition:
```java
status == null
                || (status.equals(TaskStatus.ENQUEUED) || status.equals(TaskStatus.PROCESSING))
```

the-sinner:
# Pull Request

## Related issue
Fixes #725

## What does this PR do?
- fix WaitForTask timeout in case of a task cancellation

## PR checklist

### DIFF
--- a/src/main/java/com/meilisearch/sdk/TasksHandler.java
+++ b/src/main/java/com/meilisearch/sdk/TasksHandler.java
@@ -143,7 +143,7 @@ public class TasksHandler {
         long elapsedTime = 0;
 
         while (status == null
-                || (!status.equals(TaskStatus.SUCCEEDED) && !status.equals(TaskStatus.FAILED))) {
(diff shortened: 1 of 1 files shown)


curquiza (approved):
thanks a lot!

bors merge

Issue #737, opened 2024-05-21
(JsonHander) targetClass parameter should be of type Class<T>, not Class<?>


Mashull:
**Description**

The targetClass parameter should be of type Class<T>, not Class<?>. This provides stronger type safety because you can check the type of T at compile time rather than at run time

curquiza:
closed by #738

PR #738, opened 2024-05-22
Adjust the targetClass parameter variable of the decode method in JsonHandler from class<?> to class<T>

Mashull:
## Related issue
Refactor #737

## What does this PR do?
The targetClass parameter should be of type Class<T>, not Class<?>. This provides stronger type safety because you can check the type of T at compile time rather than at run time


curquiza (approved):
thank you!

bors merge

PR #757, opened 2024-07-10
Fixing empty TaskError response

larskristianhaga:
# Pull Request

Seems like the `TaskError` object is only returning empty Strings when trying to get the error details of a task. Seems to be a missmatch between the field naming, and number of fields based on the API response. Therefor it dosent understand what to map, and only gives empty strings back.

## Related issue
None

## What does this PR do?
- It moves the `TaskError` class to the `model` package, closer to the usage in `Task` class.
  - Since its peers`TaskDetails`, and `TaskStatus` are already here.
- Updates the field in `TaskError` to match the API naming, and number of fields.
- Updates the Exception integration test to use the correct naming and import.
- Adds Integration tests to check that `TaskError` fields is not blank or null.

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/src/main/java/com/meilisearch/sdk/model/Task.java
+++ b/src/main/java/com/meilisearch/sdk/model/Task.java
@@ -1,6 +1,5 @@
 package com.meilisearch.sdk.model;
 
-import com.meilisearch.sdk.TaskError;
 import java.util.Date;
 import lombok.Getter;
 
--- a/src/main/java/com/meilisearch/sdk/model/TaskError.java
+++ b/src/main/java/com/meilisearch/sdk/model/TaskError.java
@@ -1,4 +1,4 @@
-package com.meilisearch.sdk;
+package com.meilisearch.sdk.model;
 
 import lombok.Getter;
 import lombok.Setter;
--- a/src/main/java/com/meilisearch/sdk/model/TaskError.java
+++ b/src/main/java/com/meilisearch/sdk/model/TaskError.java
@@ -3,14 +3,15 @@ package com.meilisearch.sdk.model;
 import lombok.Getter;
 import lombok.Setter;
 
-/** The code, type and error of the task error */
+/** The code, type‚ error and message of the task error */
 @Getter
 @Setter
 public class TaskError {
 
     public TaskError() {}
 
-    protected String taskErrorCode = "";
-    protected String taskErrorType = "";
-    protected String taskErrorLink = "";
+    protected String code = "";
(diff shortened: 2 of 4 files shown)


larskristianhaga:
Adding screenshot to explain more

Screenshot of the Tasks list in Meilisearch:
screenshotMeilisearchTaskList

Basic code using the SDK to get a task:
screenshotCode

Screenshot of the debugger, showing the `TaskError` object with empty String values:
screenshotDebugger

curquiza:
Hello @larskristianhaga 
thanks for your PR!

Can you confirm your changes are not breaking for the usage?
I have no issue releasing breaking in this package, I just want to ensure clear communication around it

larskristianhaga:
I would say that the changes are a breaking change, due to the renaming of the fields. But only if you actually use the `TaskError` class.
Makes more sense to include this in the next MINOR change. Eg when `0.13.0` comes out, and make a note of it in the changenotes.

I've also just pushed some integration tests to actually test the changes in the PR, and verify they have a value.

curquiza:
Thank you @larskristianhaga 
can you explain clearly what the users had to do before, and what they have to do now (after your PR). I need this for the changelog

curquiza:
It's not obvious for me where is the breaking. For me
- before users had "TaskError" but could not manipulate it
- now they still have "TaskError" but could use methods in it

What is the naming breaking you mentioned?

> I would say that the changes are a breaking change, due to the renaming of the fields

larskristianhaga:
The breaking changes that applies to this PR:

* TaskError class is moved to the model package, meaning the import will need to be changed.

Old import:
```
import com.meilisearch.sdk.TaskError;
```
New import:
```
import com.meilisearch.sdk.model.TaskError;
```

* Renaming all fields inside the TaskError class, to match what the API actually returns, so the fields will be populated.

Old code:
```
TaskError taskError = taskResult.getError();

taskError.getTaskErrorCode();
taskError.getTaskErrorType();
taskError.getTaskErrorLink();
```

New code
```
TaskError taskError = taskResult.getError();

taskError.getCode();
taskError.getType();
taskError.getLink();
```

Other changes:
* Adding a new field called message to the TaskError class.

curquiza (approved):
Thank you @larskristianhaga 

bors merge

PR #835, opened 2025-04-04
Add check for empty API key before adding Authorization header

### LINKED ISSUE #834: CustomOkHttpClient uses an empty API key from Config to set Authotization Header
**Description**
When creating a client with the host-only Config constructor, an error about wrong API key is shown even if Meilisearch instance doesn't enforce the use of API key

**Expected behavior**
The client should work if Meilisearch instance doesn't enforce the use of API key

**Current behavior**
CustomOkHttpClient looks for `null` only and not empty string to decide if it needs to add the `Authorization` header

**Affected code**
Config: https://github.com/meilisearch/meilisearch-java/blob/main/src/main/java/com/meilisearch/sdk/Config.java#L28-L30
CustomOkHttpClient: https://github.com/meilisearch/meilisearch-java/blob/main/src/main/java/com/meilisearch/sdk/http/CustomOkHttpClient.java#L59-L60

**Environment (please complete the following information):**
- OS: [e.g. Debian GNU/Linux]
- Meilisearch version: [e.g. v.0.20.0]
- meilisearch-java version: [e.g v0.3.1]

svavassori:
# Pull Request

## Related issue
Fixes #834

## What does this PR do?
- It adds empty string check before adding the `Authorization` header, so to make it consistent with the default apiKey value set by the host-only Config constructor

## PR checklist

### DIFF
--- a/src/main/java/com/meilisearch/sdk/http/CustomOkHttpClient.java
+++ b/src/main/java/com/meilisearch/sdk/http/CustomOkHttpClient.java
@@ -56,7 +56,7 @@ public class CustomOkHttpClient {
         Request.Builder builder = new Request.Builder();
         builder.url(url);
 
-        if (this.config.getApiKey() != null)
+        if (this.config.getApiKey() != null && !this.config.getApiKey().isEmpty())
             builder.addHeader("Authorization", this.config.getBearerApiKey());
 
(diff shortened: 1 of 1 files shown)


svavassori:
@curquiza any news or doubts about this PR?

curquiza (approved):
bors merge

PR #840, opened 2025-04-19
Add backwards-compatible serialization for filtration

Noah-Mack-01:
### New Additions
* **`FilterableAttribute`** is a flattened POJO implementation of the new [granular filtration feature](https://github.com/meilisearch/meilisearch/issues/5163). It contains backwards compatibility with legacy data.
* **`GsonFilterableAttributeSerializer`** allows for the serialization/deserialization of these POJOs from legacy primitive string filter attributes, as well as the persistence of new single-string attributes with default settings in legacy format. ### Changes
* **`GsonJsonHandler`** registers our new serializer.
* **`Index`**: `getFilterableAttributesSettings` is updated to accommodate the new POJO. Backwards compatible `legacyGetFilterableAttributesSettings` which returns the traditional `String[]` attribute configuration has been added as well. `updateFilterableAttributes` has been updated to accept `Object[]`, as polymorphic backwards compatibility which can accept `null` values, requires no ambiguity; as we cannot do a polymorphic method, this seemed the best fit.
* **`Settings`**: `filterableAttributes` was updated to now be of type `FilterableAttribute[]`.
* **`SettingsHandler`**: `updateFilterableAttributesSettings` now accepts a `FilterableAttribute[]` object. `getFilterableAttributesSettings` now returns one likewise.

# Pull Request

## Related issue
Fixes #<issue_number>

## What does this PR do?
- ...

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -227,7 +227,10 @@ get_filterable_attributes_1: |-
   client.index("movies").getFilterableAttributesSettings();
 update_filterable_attributes_1: |-
   Settings settings = new Settings();
-  settings.setFilterableAttributes(new String[] {"genres", "director"});
+  Map<String, Boolean> filtersTypes = new HashMap<>();
+  filtersTypes.put("comparison",true);
+  filtersTypes.put("equality",true);
+  settings.setFilterableAttributes(new FilterableAttributes[] {new FilterableAttributes("genres"), new FilterableAttributes(new String[]{"director"}, true, filters)});
   client.index("movies").updateSettings(settings);
 reset_filterable_attributes_1: |-
   client.index("movies").resetFilterableAttributesSettings();
--- a/src/main/java/com/meilisearch/sdk/Index.java
+++ b/src/main/java/com/meilisearch/sdk/Index.java
@@ -5,6 +5,7 @@ import com.meilisearch.sdk.http.URLBuilder;
 import com.meilisearch.sdk.model.*;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import lombok.Getter;
@@ -443,8 +444,8 @@ public class Index implements Serializable {
      * @see <a
      *     href="https://www.meilisearch.com/docs/reference/api/facet_search#perform-a-facet-search">API
      *     specification</a>
-     * @see Index#getFilterableAttributesSettings() getFilterableAttributesSettings
-     * @see Index#updateFilterableAttributesSettings(String[]) updateFilterableAttributesSettings
+     * @see Index#legacyGetFilterableAttributesSettings() getFilterableAttributesSettings
+     * @see Index#updateFilterableAttributesSettings(Object[]) updateFilterableAttributesSettings
      * @since 1.3
      */
     public FacetSearchable facetSearch(FacetSearchRequest facetSearchRequest)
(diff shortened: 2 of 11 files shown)


Noah-Mack-01:
@Strift  PR in relation to #838 

LMK if you have any questions or concerns!

.code-samples.meilisearch.yaml
```diff
@@ -227,7 +227,10 @@ get_filterable_attributes_1: |-
   client.index("movies").getFilterableAttributesSettings();
 update_filterable_attributes_1: |-
   Settings settings = new Settings();
-  settings.setFilterableAttributes(new String[] {"genres", "director"});
+  Map<String, Boolean> filtersTypes = new HashMap<>();
+  filtersTypes.put("comparison",true);
+  filtersTypes.put("equality",true);
+  settings.setFilterableAttributes(new FilterableAttributes[] {new FilterableAttributes("genres"), new FilterableAttributes(new String[]{"director"}, true, filters)});
```
Strift, line 233:
```suggestion
  settings.setFilterableAttributes(new FilterableAttribute[] {new FilterableAttribute("genres"), new FilterableAttribute(new String[]{"director"}, true, filters)});
```

curquiza:
Hello!

Thank you for your PR @Noah-Mack-01! ❤️ 

If I understand correctly, this PR introduces breaking changes, because people need to use the `legacy` methods to have the same behavior than before.
It's ok to make breaking changes, we are not stable yet in this library, but we need to know which ones to do the appropriate changelogs

Also, is there a way to avoid breaking change? Make the new methods to accept both of format (array of string and array of object)?

I would love having the review of @brunoocasali before merging that

Noah-Mack-01:
Hi @curquiza !!

Thinking through this now. Its unfortunate we're working off such an old JDK, since Java 15's `sealed` interface would be useful in this situation. There's only one method that's been migrated to a `legacy` tag, and thats a getter method to retrieve the older String[] format of filters. 

All update methods are generic, such as the following:
`    public <O> TaskInfo updateFilterableAttributesSettings(O[] filterableAttributes)
            throws MeilisearchException {
        if (filterableAttributes == null)
            return this.settingsHandler.updateFilterableAttributesSettings(this.uid, null);
        else if (filterableAttributes.getClass().getComponentType() == FilterableAttribute.class)
            return this.settingsHandler.updateFilterableAttributesSettings(
                    this.uid, (FilterableAttribute[]) filterableAttributes);
        else if (filterableAttributes.getClass().getComponentType() == String.class)
            return this.updateFilterableAttributeSettingsLegacy((String[]) filterableAttributes);
        else
            throw new MeilisearchException(
                    "filterableAttributes must be of type String or FilterableAttribute");
    }

    private TaskInfo updateFilterableAttributeSettingsLegacy(String[] filterableAttributes) {
        return this.settingsHandler.updateFilterableAttributesSettings(
                this.uid,
                Arrays.stream(filterableAttributes)
                        .map(FilterableAttribute::new)
                        .toArray(FilterableAttribute[]::new));
    }`
 
 However I can't think of a way to guarantee that the `getFilterableAttributesSettings()` method will return EITHER a String[] or FilterableAttribute[], and we can't utilize polymorphism here since there aren't parameters to the method.
 
 I suppose the answer would be to maintain the legacy implementation's original nomenclature `getFilterableAttributeSettings` and refer to the new method's implementation as something else. Let me know what you think!!!

Noah-Mack-01:
Hi @curquiza @Strift 

Made changes I described above. I think the compiler's inability to discern method by result class will be a hard limitation for the current implementation of `Index`. Please LMK if there's anything else you think would be prudent to change!

src/main/java/com/meilisearch/sdk/Index.java
```diff
@@ -747,24 +748,67 @@ public TaskInfo resetLocalizedAttributesSettings() throws MeilisearchException {
...
+    }
+
+    private TaskInfo updateFilterableAttributeSettingsLegacy(String[] filterableAttributes) {
         return this.settingsHandler.updateFilterableAttributesSettings(
-                this.uid, filterableAttributes);
+                this.uid,
+                Arrays.stream(filterableAttributes)
+                        .map(FilterableAttribute::new)
+                        .toArray(FilterableAttribute[]::new));
     }
```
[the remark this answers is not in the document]
Noah-Mack-01: This is not a public method and its references in `Index.java` appropriately handle the new parameters, including with backwards compatibility for the original String[] inputs... Don't think this revision is necessary.

src/test/java/com/meilisearch/integration/ClientTest.java
Noah-Mack-01, line 1: This fails bc of Docker image upgrade to Java 11 
Broadened expected error for now -- existing TODO covers what needs to be implemented

brunoocasali (changes requested):
Hey @Noah-Mack-01 thanks a lot for sending this PR!

I proposed a simpler API for sending the new filterable attributes, tell me what do you think and if it is possible to do. I really believe if possible will make the life of the users easier, especially because of the rest of the SDK which has the same style.

Then we can talk about the underlying question regarding the compiler/method names and etc...

